PR #3, opened 2015-08-26
regression: zonal setting missing

mdzik:


llaniewski:
This issue arose from malfunctioning "indent" tool. Indent is correcting the indentation (tabs) and the general "style" of the code. But it looks like it does weird things with comments sometimes. The indent was switched off and this whole part was deleted from the acGeometry. Everything is working now without it.

https://github.com/llaniewski/TCLB/blob/develop/src/Handlers.cpp.Rt#L2000

mdzik:
last mainstream version is ok, sorry

PR #7, opened 2015-08-28
extending pythoncall features :)

mdzik:
extending to multiple Densities support, only that makes sense :)
fixing PYTHONPATH to support any external library, like mpi4py

it *should* be possible to write native C extensions, wrapped to CLB via python using blitz

and now you could call TCLB "The Monster" ;)


llaniewski:
I'll keep this open until you're finished :-)

mdzik:
ok, but i think i'm out of ideas for now :-)

mdzik:
...something is definetly wrong with commits/merges...

llaniewski:
pushing to TCLB is like catching a riding train :-) tedious, but fun.

llaniewski:
If you want to push a commit, without running tests on travis, add [ci skip] to commit message

mdzik:
Now I know, but this should compile. After all this conflicts :p

mdzik:
outdated

PR #15, opened 2016-01-07
Latest changes to cumulant model,averaging,etc

shkodm:


src/conf.R
```diff
@@ -177,6 +179,7 @@ AddSetting = function(name,  comment, default=0, unit="1", adjoint=F, derived, e
 	} else {
 		if (missing(equation)) stop("'derived' provided, but no 'equation' in AddSetting")
 	}
+	if (name == 'Ave') Ave <<-default;
```
llaniewski: You don't need to go through here to modify a global variable.

You can just have a "Ave = TRUE" somewhere in the beginning of Dynamics.R

src/d3q27_cumulant/Dynamics.R
```diff
@@ -15,11 +15,13 @@ AddDensity(
...
 AddQuantity( name="U",unit="m/s",vector=T)
 
-AddSetting(name="nu", default=1.6666666, comment='Viscosity',zonal=TRUE)
+AddSetting(name="nu", default=1.6666666, comment='Viscosity')
+AddSetting(name="nubuffer",default=0.01, comment='Viscosity in the buffer layer')
 AddSetting(name="Velocity", default="0m/s", comment='Inlet velocity', zonal=TRUE)
 AddSetting(name="Pressure", default="0Pa", comment='Inlet pressure', zonal=TRUE)
 AddSetting(name="Smag", comment='Smagorinsky constant')
 AddSetting(name="Turbulence", comment='Turbulence intensity', zonal=TRUE)
+AddSetting(name="Ave",default=0,comment='Averaging indicator') #Change default to 1,if you want for average values to be calculated,0 - for default,fast solution
```
llaniewski: Ave = TRUE is enough

src/d3q27_cumulant/Dynamics.R
```diff
@@ -32,3 +34,30 @@ AddNodeType("Smagorinsky", "LES")
 AddNodeType("Stab", "ENTROPIC")
 AddNodeType("SymmetryY", "BOUNDARY")
 AddNodeType("SymmetryZ", "BOUNDARY")
+AddNodeType("Tinlet","BOUNDARY")
```
llaniewski: Make it WVelocityTurbulent for now. But at the end, I think we don't need a separate boundary node type for that.

src/d3q27_cumulant/Dynamics.R
```diff
@@ -32,3 +34,30 @@ AddNodeType("Smagorinsky", "LES")
 AddNodeType("Stab", "ENTROPIC")
```
llaniewski: I think this is not needed in this model

src/d3q27_cumulant/Dynamics.c.Rt
```diff
@@ -121,17 +166,38 @@ CudaDeviceFunction void SetEquilibrum(real_t rho, real_t Jx, real_t Jy, real_t J
 }
 
 CudaDeviceFunction void Init() {
+	real_t k_aa = exp(-1/constContainer.ST.TimeWN);
+        real_t k_bb =  sqrt(1 - k_aa*k_aa);
 	vector_t ST = SyntheticTurbulence(0);
-	ST.x = Velocity + Turbulence*ST.x;
-	ST.y = Turbulence*ST.y;
-	ST.z = Turbulence*ST.z;
+	ST.x = Turbulence*ST.x*k_bb + SynthTX*k_aa;
```
llaniewski: This doesn't make sens in Init. Init should not use time correlation

src/d3q27_cumulant/Dynamics.c.Rt
```diff
@@ -121,17 +166,38 @@ CudaDeviceFunction void SetEquilibrum(real_t rho, real_t Jx, real_t Jy, real_t J
...
+	SynthTY = ST.y;
+	SynthTZ = ST.z;
+	ST.x = ST.x + Velocity;	
 	SetEquilibrum(1.0 + Pressure * 3.0, ST.x, ST.y, ST.z);
 }
 
+CudaDeviceFunction void SynthInit() {
+        real_t k_aa = exp(-1/constContainer.ST.TimeWN);
+        real_t k_bb =  sqrt(1 - k_aa*k_aa);
+       	vector_t ST = SyntheticTurbulence(0);
```
llaniewski: ST won't be visible in the boundary condition implementation, because it's local to SynthInit.

llaniewski:
@shkodm Please fix the above issues before we continue with the pull request.

shkodm:
I've fixed almost all issues,except for boundary condition on the synthetic turbulence nodes. I will push everything  once I do it also,hopefully tonight.

llaniewski:
@shkodm Merged. Please get up to date with the upstream.

PR #17, opened 2016-01-14
Adding working force term,boundary condtion and galilean correction

shkodm:


llaniewski:
@shkodm Please add half a force term to getU.

PR #24, opened 2016-02-04
added suport for < and > in dx and nx

mdzik:


src/Geometry.cpp.Rt
```diff
@@ -138,27 +160,46 @@ lbRegion Geometry::getRegion(const pugi::xml_node & node)
...
+    char side; 
+    
     attr = node.attribute("dx");
     if (attr) {
-	int w = val(attr);
-	if (w < 0)
-	    w = ret.nx + w;
+	int w = val_p(attr, &side);
+    
+    if (side == '<') {
```
llaniewski: @mdzik I think we want:
< from left side
> from right side
+ old style

Now it is opposite. Please correct.

PR #27, opened 2016-03-09
Syntax files for vim&tclb

mdzik:
[ci skip]


mdzik:
...and ci skip seems not to work in comments ;)

llaniewski:
It's [ci:skip] and I think it have to be in the commit msg.

PR #28, opened 2016-03-21
Fixing double/single precision config

shkodm:


mdzik:
Maybe we will consider setting up double precision/cuda sm_25(?) as default? ("if you don't know what you are doing, use double" approach)

llaniewski:
@mdzik is right. @shkodm please make it
'not equal to "xno"'
In place of
 'equal to "xyes"'

Otherwise float will be default (as default in configure is "")

mdzik:
it's so nice to be right :laughing: 

For the future: maybe there is some better way to set defaults in AC? maybe at the beginning, where you declare params? That would be easier to maintain (all defaults in one place, not hidden in 'not equal to...' everywhere)

PR #29, opened 2016-03-21
Added OffGrid Sphere geometry element

mdzik:
Added OffGrid circle element: it does make difference for multiple-circle geometry
+example


src/Geometry.cpp.Rt
```diff
@@ -490,8 +490,26 @@ int Geometry::Draw(pugi::xml_node & node)
...
-	}    
-    else if (strcmp(n.name(),"Pipe") == 0) {
+    } else if (strcmp(n.name(), "OffgridSphere") == 0) {
+
+        double x0 = units.alt(n.attribute("x").value());
+        double y0 = units.alt(n.attribute("y").value());
+        double z0 = units.alt(n.attribute("z").value());
+        double R0 = units.alt(n.attribute("R").value());
+        double R2 = R0*R0;
+	    debug1("Filling sphere with flag %d (%d)\n, with center at (%f,%f,%f) and radius %f", fg, fg_mask, x0, y0, z0, R0);
```
llaniewski: %lf not %f

src/Geometry.cpp.Rt
```diff
@@ -490,8 +490,26 @@ int Geometry::Draw(pugi::xml_node & node)
...
 			}
 		    }
-	}    
-    else if (strcmp(n.name(),"Pipe") == 0) {
+    } else if (strcmp(n.name(), "OffgridSphere") == 0) {
+
+        double x0 = units.alt(n.attribute("x").value());
+        double y0 = units.alt(n.attribute("y").value());
+        double z0 = units.alt(n.attribute("z").value());
+        double R0 = units.alt(n.attribute("R").value());
```
llaniewski: I think there should be 3 R's, and by default "R" will overwrite them all.

src/Geometry.cpp.Rt
```diff
@@ -490,8 +490,26 @@ int Geometry::Draw(pugi::xml_node & node)
...
+        double R2 = R0*R0;
+	    debug1("Filling sphere with flag %d (%d)\n, with center at (%f,%f,%f) and radius %f", fg, fg_mask, x0, y0, z0, R0);
+	    DEBUG1(reg.print();)
+		for (int x = reg.dx; x < reg.dx + reg.nx; x++)
```
llaniewski: the "reg" should be and intersection of the inferred region bounding the sphere and the surrounding region. That way you will not have to provide dx,dy,nx etc at all.

PR #36, opened 2016-03-30
Templates for ISSUE/PULL REQUEST

mdzik:
proposal to resolve #34

Issue #34, opened 2016-03-29
Merge policy


mdzik:
New rule proposal:
ANY new, model-related future/helper (auto BC, feq/collision helper etc.) will not be merged without reference to wiki (in code) OR at minimum meaningful API doc-string. Things like lib/boundary.R are unusable without that
 

Also, maybe we should come out with some merging guide, especially if we want someone to use it ;)

llaniewski:
See: https://github.com/blog/2111-issue-and-pull-request-templates
Draft a proposal, then we can iterate.

mdzik:
In pull request #44

mdzik:
Should we close that one?

PR #37, opened 2016-03-30
D2q9 inc

mdzik:
**Proposal**
"Incopressible" d2q9-mrt,
Feq=wi (rho + rho0*(u^2 + etc)

For low Re and Ma doesn’t seem to improve anything
Stored drho instead of rho -> better accuracy for low Ma


src/d2q9_inc/Dynamics.R
```diff
@@ -0,0 +1,50 @@
+# Density - table of variables of LB Node to stream
```
llaniewski: I know that we don't have such a thing in other models (to be filled), but it's good time to start:

Add something like "# Maintainer: Michal Dzikowski @mdzik" So we know who's model is it (as we really get lot of models now.

We should include this in #34 #36

src/d2q9_inc/conf.mk
```diff
@@ -0,0 +1,2 @@
+ADJOINT=0
+TEST=TRUE
```
llaniewski: I think we can exclude this from tests - I'm planning to reduce the number of tested models (or make some different policy for testing) in near future.

llaniewski:
Is this still alive, or is this obsolete wrt #40 ?

mdzik:
Different, aims to be "better" for low-low-speed calculations, rho - 1 trick etc. Nothing to do with #40 
It's working, but I can't see any significant difference (Ladd says, that rho-1 trick is rather important for ~creping flows, even in double precision)
Works & mergable, but probably to extend

PR #39, opened 2016-04-01
Adding Sampler class

shkodm:
Sampler class allows to record and store quantities  from each iteration in a separate GPU buffer. Output can then be done in any convenient time.


src/Handlers.cpp.Rt
```diff
@@ -923,6 +926,57 @@ class cbFailcheck: public Callback {
...
+		char fn[STRING_LEN];
+		Callback::Init();
+		pugi::xml_attribute attr=node.attribute("what");
+		if (attr) {
+			s.add_from_string(attr.value(),',');
+		}
+		else {
+			s.add_from_string("all",',');
+		}
+		for (pugi::xml_node par = node.first_child(); par; par = par.next_sibling()) {
```
llaniewski: Please check if the child is a "Point" - otherwise error.

src/Lattice.cu.Rt
```diff
@@ -444,7 +445,9 @@ inline void Lattice::<?%s FunName ?>(int tab0, int tab1, int iter_type)
 <?R if (stage$fixedPoint) { ?> } // for(fix) <?R } ?>
 <?R } ?>
 	Snap = tab1;
+	lbRegion damn;
```
llaniewski: Is this used anywhere?

src/Handlers.cpp.Rt
```diff
@@ -2004,7 +2062,33 @@ class acSyntheticTurbulence: public Action {
 			}
 			output("Setting one-wave spectrum with length=%lg\n", 1/WN); 
 			solver->lattice->ST.setOneWave(WN);
-		} else {
+		}/* else if (spec == "Free stream") {
```
llaniewski: Let's get rid of this commented out section. I think "Intensity" can be added directly to the "Von Karman" spectrum as an option - but that's a matter for another pull request.

PR #40, opened 2016-04-06
D2q9 plain phase-field model  and new multiphase model

mdzik:


example/phase-field.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0"?>
+<CLBConfig version="2.0">
+
+
+    <!--- WARNING -->
```
llaniewski: I think you can get rid of this.

PR #45, opened 2016-04-11
Adding a special copy with preserve the filename for coveralls

### LINKED ISSUE #31: Coveralls not working - no source to look at
The file "CLB/d2q9/Geometry.cpp" isn't available on github. Either it's been removed, or the repo root directory needs to be updated.

llaniewski:
Should resolve #31


llaniewski:
#31 nearly solved. Only Geometry.cpp.Rt is wrong.

Issue #49, opened 2016-04-28
Execution after Failcheck finds NAN


llaniewski:
It would be nice if we could execute something still after the failcheck finds NANs.

I propose this syntax:
```xml
<Failcheck>
  <VTK name="withnan" what="ALL"/>
</Failcheck>
```

This would write the VTK file after finding the nan. This could help identify that is happening better. I thinks it can be done with something like ExecuteInternal().

shkodm:
Yea, it is really good idea. I will implement it, but the "syntax" you proposed is not visible...

llaniewski:
@shkodm I fixed the syntax

llaniewski:
Closed by #54

PR #54, opened 2016-05-05
Execution after NaN

shkodm:
I implemented, what was proposed in issue #49 . Now post-Failcheck action can be performed, such as basic VTK or even something like this:
```xml
<Failcheck Iterations="1000">
         <Solve Iterations="2000">
               <VTK name="withnan" Iterations="500"/>
         </Solve>
</Failcheck>
```

### DIFF
--- a/src/Handlers.cpp.Rt
+++ b/src/Handlers.cpp.Rt
@@ -890,12 +890,14 @@ class cbAveraging: public Callback{
 
 class cbFailcheck: public Callback {
 	lbRegion reg;
+	int rkept;
 	public:
 	int Init() {
 		Callback::Init();
 		reg.dx = solver->region.dx;
 		reg.dy = solver->region.dy;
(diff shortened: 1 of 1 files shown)


src/Handlers.cpp.Rt
```diff
@@ -916,10 +918,15 @@ class cbFailcheck: public Callback {
 		}
 		<?R }; ifdef(); ?>
 		MPI_Allreduce(&cond,&fin,1,MPI_INT,MPI_LOR,MPI_COMM_WORLD);
-		if (fin) { 
-			 notice("Stopping due to Nan value\n");
-			 ret = ITERATION_STOP;
-		}
+	        if ((fin) && (rkept)) {
+                         rkept = 0;
```
llaniewski: Make a "notice" also here. something like "Nan value discovered. Executing the final wishes from the FailCheck element." :-)

PR #58, opened 2016-06-02
Adding B-Spline sphere sweep

llaniewski:
I added a geometry element `<Sweep>`, which does a sphere sweep along a B-Spline defined by points. It's a bit slow - but works nicely

Example:
```xml
<Sweep order="2" r="10" steps="300">
<Point x="20" y="50"/>
<Point x="50" y="50"/>
<Point x="70" y="30"/>
<Point x="90" y="30" r="20"/>
</Sweep>
```


llaniewski:
Also added ZouHeRewrite - but it's not yet used anywhere.

llaniewski:
@shkodm You can use this [B-Spline function](https://github.com/CFD-GO/TCLB/blob/develop/src/spline.h#L37) in your work - it's a standalone header. You could use this sphere sweep also, but it is very slow for now.

PR #60, opened 2016-06-06
Adding support for "drawing" mode in Geometry

llaniewski:
I added the feature in the Geometry module, that allows to specify if a node should be:
- `mode="overwrite"` -- Overwrites the previous value of the node type
- `mode="change"` -- Overwrites only non-zero node types.
- `mode="fill"` -- Sets the node type only in elements that didn't have the node type earlier (had 0 in this `mask`)

Example:
```xml
<EVelocity mode="fill">
```
will set the EVelocity boundary condition only on elements that don't have previously set boundary condition.

@mdzik @shkodm

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -14,6 +14,8 @@ We are really glad that you would like to contribute to TCLB. We would appreciat
 - [ ] You did a clever trick? Please leave a suitable comment
 - [ ] Please don't brake naming convention (at least look at surrounding code)
 - [ ] Try too use meaningful variable names
+- [ ]  Do not pull/merge XXX_new's.  Its either something different and should have different name or it is an replacement of XXX (in whole code)
+
 
 ### Compilation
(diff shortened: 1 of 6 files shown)


mdzik:
What about zones? They go by the same rule?

llaniewski:
@mdzik The rule applies to all things covered by the "mask". So for instance `<EVelocity mode="fill" name="Inlet">` will not modify elements that have a zone name already set.

I was considering adding an additional mask for the `mode` feature, but I didn't want to overcomplicate it.

llaniewski:
I fixed the merge conflict. @mdzik I will merge after the test success. We can modify the behaviour later if we have better ideas how it should work.

@mdzik @shkodm The default `mode` is "overwrite" so if you don't specify it, it will work as earlier.

mdzik:
OK.
So if I have <None name="test"> overlapped  by <EVelocity name="test2"> then ??
Maybe picture in wiki?
This behaviour need to be clearified

PR #61, opened 2016-06-14
Adding proper Symmetry BC

shkodm:
I added properly working template for symmetry boundary condition. 
First argument is direction of symmetry, second - sign of vector; similarly to ZouHe. 

Example: 
```cuda
CudaDeviceFunction void BottomSymmetry() {
<?R SymmetryNew(2, 1) ?>
}

### DIFF
--- a/src/Handlers.cpp.Rt
+++ b/src/Handlers.cpp.Rt
@@ -1833,7 +1833,7 @@ class conControl: public Action {
                         int k = 0;
                         int max_k = csv_data[ "_time" ].size() - 2;
                         if (max_k < 1) {
-                                error("Not enaugh records in CSV file or something went terribly wrong\n");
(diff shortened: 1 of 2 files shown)


mdzik:
OK, so if old one is "non properly" working, pleas remove it, rename SymmetryNew to Symmetry and replace in requiring models

mdzik:
And one general comment: **pleas do not pull/merge XXX_new's**, its either something different and have different name or should replace XXX in all codebase. What would you do, when you decide to improve XXX_new? it will be XXX_new_new or XXX_verynew?

In case of models: new implementation starts its life in experimental/ then replace in stable/

shkodm:
@mdzik, that is really good idea. For ZouHe, we already have ZouHe, ZouHeNew and ZouHeRewrite...

mdzik:
Pull request #62

PR #64, opened 2016-06-15
Substitution to new Symmetry BC 

shkodm:
Adding modifications, as discussed in #61


src/d3q27/Dynamics.R
```diff
@@ -28,5 +28,5 @@ AddGlobal(name="Flux", comment='Volume flux', unit="m3/s")
 
 AddNodeType("Smagorinsky", "LES")
 AddNodeType("Stab", "ENTROPIC")
-AddNodeType("SymmetryY", "BOUNDARY")
-AddNodeType("SymmetryZ", "BOUNDARY")
+AddNodeType("SymmetryYTop", "BOUNDARY")
+AddNodeType("SymmetryZIn", "BOUNDARY")
```
mdzik: Change SymmetryXXX to N/S/W/E/I/OSymmetry -> like EPressure  (?)

mdzik:
Does it work with 2D models?

mdzik:
:+1:

shkodm:
It's idea is similar to Bounceback function. It works, if  `Density$group`  is specified when declaring `f` declaration. For 2D, 3D, whatever. So it works fine for the d2q9_kuper for instance, but not for d2q9.

PR #81, opened 2016-06-30
Update CONTRIBUTING.md

shkodm:
More info and, alignment and  spelling correction.


mdzik:
OK, but you need to rephrase that: 
"When working with model independent kernel files and handlers, take a look at the surrounding code. A lot of classes and useful function are already implemented. Remember: 'grep' is your friend "

This basically means: we do not have ANY meaningful dev-doc, so if YOU want US to consider your input, you need to first read whole codebase, looking for clues about nice and fancy solutions hidden between old and grumpy ones. There are a lot of projects out there that died because of that attitude. 

Coding standard are rules/requirements, maybe something like that could be moved to tips or sth?

If we want to enforce "standard solutions of standard problems" they should be state in wiki/doc/example and this one should be taken as *must read* before contributing.

shkodm:
You are right. It should be a bit rephrased, and placed in "TIPS" section.

PR #83, opened 2016-07-07
PythonInline geometry element

mdzik:
I added the PythonInline geometry element. Bit experimental*, but with this making "fancy" geometry is much cleaner: no need to outside script.
Performance should be comparable to the rest of the code, function is parsed/compiled once and then executed.
Additional you have state-full  Python environment in whole "< nx= ny=>" box

*) might be unstable, I hope I fixed all thread/gis related issues, but one never knows. Should have no impact on solving part of the solver. Python is finalized after processing geometry block.

### DIFF
--- a/src/Solver.cpp.Rt
+++ b/src/Solver.cpp.Rt
@@ -573,7 +573,9 @@ int Solver::getQuantitieIntoBuffer(const char* comp, real_t* &buf, long int*  di
         <?R if (q$type == 'vector_t') { ?>
             dim[3] = 3;
         <?R } ?>
+        #ifdef ADJOINT
     	lattice->Get<?%s q$name ?>(region, tmp_<?%s q$name ?>, 1);
+        #endif
         buf = (real_t*)tmp_<?%s q$name ?>;
         somethingWritten = true; 
     }
--- a/src/Solver.cpp.Rt
+++ b/src/Solver.cpp.Rt
@@ -566,21 +566,20 @@ int Solver::getQuantitieIntoBuffer(const char* comp, real_t* &buf, long int*  di
 	output("Providing quantitie %s to buffer..\n", comp);
     bool somethingWritten = false;
 <?R 
-    for (q in rows(Quantities)) { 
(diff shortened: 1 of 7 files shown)


src/Solver.cpp.Rt
```diff
@@ -543,13 +543,61 @@ int Solver::getComponentIntoBuffer(const char* comp, real_t* &buf, long int*  di
...
+
+    
+    offsets[0] = region.dx;
+    offsets[1] = region.dy;
+    offsets[2] = region.dz;
+
+	output("Providing quantitie %s to buffer..\n", comp);
+    bool somethingWritten = false;
+<?R 
+    for (q in rows(Quantities)) {
```
llaniewski: you need some ifdef Adjoint here. like in other places with Quantities.

mdzik:
ifdef should "protect" against calling Solver->Get_q$name ? But preserve memory allocation etc.?

static_cast: http://stackoverflow.com/a/22330938

llaniewski:
ifdef should switch off parts of code that are adjoint specific. There is a helper function ifdef in R, which takes a bool and injects the `#ifdef ADJOINT`, you can see the usage here:

https://github.com/CFD-GO/TCLB/blob/a5885da3c6d67d2851da5ffa0a94ef1729d31873/src/Catalyst.cpp.Rt#L126

mdzik:
```c
      <?R for (q in rows(Quantities)) { ifdef(q$adjoint); ?>
/////////////////////////////////////////////
      <?R
	};
	ifdef();
?>
```
why closing ifdef(); is after R for loop?

llaniewski:
because it have to close the eventual opened `#ifdef`'s

mdzik:
I'm not that dumb ;) 
But why AFTER closing _for_ {}?
I would expect  
```R
for (q in rows(Quantities)) {
       ifdef(q$adjoint); 
#sth
       ifdef();
};
?>
```

mdzik:
ops, last one was supposed to go to separate branch.

PR #84, opened 2016-07-11
Update Dynamics.R

shkodm:
I changed Galilean correction default setting to 1(enabled), thus no need to specify it in `<Params/>` anymore. I believe, it was like that, since until fixed in #15, default settings were not working (were just setting to zero).


mdzik:
No reason to wait
@shkodm do you have time to think about test for this model? You could make test by yourself or you could give/point me to best case file. It need to be small in space and time, so TravisCI virtual machine could handle it.

shkodm:
@mdzik , sure. I wanted to add tests for this model for a long time. I will think about it and figure something out.

PR #85, opened 2016-07-15
Fix some stuff related to the d2q9_adj model in order to make them compile. Not sure if all this is correct, please check carefully.

franjesus:
The AD pre/post stuff did variable-width numbers for the 0xFFF?? variables so when there were more than 10 variables fed in to tapenade, the re-writing with sed would fail and produce variable names that were inconsistent, such as nu2 instead of PorosityGamma. I fixed it by enforcing width 2 for that int -> string conversion, which should allow up to 100 variables.

This bug was triggered because now there are additional variables such as S3, S4, S78, GravitationX... those should probably not go into tapenade, I didn't check how to disable them.

The syntax for gvector/polyAlgebra/ToC has changed and %% is no longer matrix multiplication, it is %*% instead, plus C(a == b) is no longer working, C(a,b) is the new form. Corrected Dynamics.c.Rt accordingly.

Changes to the scripts are for debugging purposes.

I am sure I did something wrong somewhere, so please review all this.


llaniewski:
@franjesus Welcome and thanks for the contribution.

As we spoke in the e-mails, I will get working on an adjoint example for you, as soon as I'll have a bit of time.

mdzik:
Hello @franjesus  :)
@llaniewski , could you send me one to? I could "convert" it into test, so we will keep adjoint on track in future

llaniewski:
@mdzik TAPENADE is proprietary. For now let's keep adjoint out of Travis-CI.

PR #86, opened 2016-07-20
Newmake

mdzik:
A little bit more portable version:
- we could not assume that CLB dir is writable nor force it
- sometime its nice to be able to have 2 build dirs, for example for CPU and GPU. Now you dont need to move them, and you could switch with configure 
- its better not to use relative paths to original source : if CLB is symlink (for example to ramdisk) this would not work

I'm open to more suggestions ;)
btw: default behavior should be preserved


mdzik:
merge?

mdzik:
obsolete

PR #90, opened 2016-07-24
Adding wall clock time to Log

llaniewski:
I added a "Walltime" column to the csv log, with the wall clock time from the beginning of execution.


tools/csvdiff
```diff
@@ -20,9 +21,17 @@ read = function(f) {
 if (! is.numeric(opt$limit)) stop("Limit not numeric.")
 eps = abs(opt$limit)
 
+opt$discard = strsplit(opt$discard,",")[[1]]
+if (! is.character(opt$discard)) stop("Distard list is not text")
```
shkodm: Typo: "Distard" should be "Discard"

PR #93, opened 2016-07-25
Fix 87

mdzik:
I fixed 87, on top fix 88


src/Handlers.cpp.Rt
```diff
@@ -2715,9 +2708,9 @@ vHandler * getHandler(pugi::xml_node node) {
 			pugi::xml_attribute attr = node.attribute("Iterations");
                 	if (attr) {
 				ret = new acSAdjoint;
-				WARNING("Makin a steady adjoint, becuase you gave me Iterations - better to state type explicitly.\n");
+				WARNING("Making a steady adjoint, becouse you gave me Iterations - better to state type explicitly.\n");
```
shkodm: It is "because", not "becouse".

PR #94, opened 2016-07-27
Fix of Sampler output

shkodm:
I fixed some issues with output format of Sampler. Now it should be in a better form and suitable for analysis.

### DIFF
--- a/src/Sampler.cpp.Rt
+++ b/src/Sampler.cpp.Rt
@@ -21,11 +21,11 @@ int Sampler::initCSV(const char *name)
      f = fopen(name, "wt");
      output("Initializing %s\n",filename);
      assert( f != NULL );
-     fprintf(f,"Iteration,X,Y,Z,");
+     fprintf(f,"Iteration,X,Y,Z");
      <?R for (q in rows(Quantities)) { ifdef(q$adjoint); ?>
      if (quant->in("<?%s q$name ?>")) 
-	<?R if (!q$vector) { ?> fprintf(f,"<?%s q$name ?>,");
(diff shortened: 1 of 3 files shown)


src/Sampler.cpp.Rt
```diff
@@ -21,7 +21,7 @@ int Sampler::initCSV(const char *name)
      f = fopen(name, "wt");
      output("Initializing %s\n",filename);
      assert( f != NULL );
-     fprintf(f,"Iteration,");
+     fprintf(f,"Iteration,X,Y,Z,");
```
llaniewski: @shkodm Please make the comma at begining of written stuff, not at the end. otherwise you'll get always a comma at the end of every line.

For instance, if you'll print "Iterations,X,Y,Z" and then write ",Rho", and ",U.x,U,y,U.z", etc. you won't get the comma at the end of line.

src/Sampler.cpp.Rt
```diff
@@ -39,10 +39,9 @@ int Sampler::writeHistory(int curr_iter) {
 		tmp_loc.x = spoints[j].location.dx;
 		tmp_loc.y = spoints[j].location.dy;
 		tmp_loc.z = spoints[j].location.dz;
-     		fprintf(f,"Point coordinates: ");
-		csvWriteElement(f,tmp_loc);
 		for (int i = startIter; i< curr_iter; i++){
```
llaniewski: @shkodm Please switch the places of the two loops - so that you'll have
for (iteration) { for (point) {write} } 
then it will be consistent throughout the file and easier to read.

PR #97, opened 2016-07-31
Fixing missing file paths

llaniewski:
Fixed missing paths for many of the optimization related files. They we're placed in the main directory always, which caused many problems for running several cases in the same directory.

plus some small fixes for TXT and sw graphic display.


llaniewski:
@mdzik Can you check what is wrong with the testing? I'll be leaving for some time and don't have time to check. When I compile on CPU at my computer, the tests do not fail.

mdzik:
Hmm, signal 11, know as "segmentation fault".
I have docker with TravisCI images up&singing, so will do

llaniewski:
@mdzik Ok. I found it. I missed one constructor.

mdzik:
God dam it, those missing constructors ;p
BTW: valgrind didn't blow on that, while plain run segfaulted. I really love that tool :/

Maybe we want to include valgrind cpu run to tests? That is simple to implement, but will hurt our feelings.

I've checked how current develop is doing in karman test. We have some memleaks, a lot of "new dim". 
Cons: Travis test will take much longer

llaniewski:
Yeah, I'm still considering Valgrind on Travis-CI, but to fix all the leaks would be very hard. This is post-PhD :D

PR #106, opened 2016-10-12
Adding Interpolated Bounce Back

llaniewski:
I added Quadratic Interpolated Bounce Back (QIBB) based on @wojtasMEiL code.
1. Added Q field to Geometry and LatticeContainer, which measures cut distances for all directions
2. Added calculation of cut distances to import of STL
3. Added "surface" mode to STL import, which sets the flag on all neighbouring nodes.
4. Added d3q27_cumulant_qibb_small model which is an extension of @shkodm cumulant model, with a QIBB boundary condition based on @wojtasMEiL side code and the equations in Geier article on Cumulant LB. (the implementation doesn't include the "reverse collision" like the original article)

You can add the QIBB geometry with:

``` xml
<QIBB>
  <STL file="example/sphere.stl" side="surface"/>
</QIBB>
```


mdzik:
Seems legit :)

PR #117, opened 2016-12-14
[V6.1 RC] Corrected the Get_ Set_ functions

llaniewski:
Corrected the Get_ Set_ and Clear_ functions in Lattice.cu, to work **only** with densities marked as `parameter`. This changes the behaviour of PythonCall, as it can now read and write only such densities/fields. You can declare such density as:
```R
AddDensity(name="G", parameter=TRUE)
```
A density which is marked as parameter can only have streaming direction 0 (dx=0,dy=0,dz=0).

### DIFF
--- a/models/flow/d2q9/Dynamics.R
+++ b/models/flow/d2q9/Dynamics.R
@@ -20,8 +20,8 @@ AddDensity( name="f[8]", dx= 1, dy=-1, group="f")
 # If present thei are used:
 # As VelocityX/Y for Boundary conditions
 # As mass force (+ GravitationX/Y) in fluid
-AddDensity( name="BC[0]", dx=0, dy=0, group="BC")
-AddDensity( name="BC[1]", dx=0, dy=0, group="BC")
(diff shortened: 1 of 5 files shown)


llaniewski:
@mdzik, we have to decide on the changes to the d2q9_kuper test, because 'f0' is not a parameter. This is the only test which fails now:
https://travis-ci.org/llaniewski/TCLB/builds/183898690

mdzik:
The  question here is if we want to be able to "alter" the real densities
(solved ones). I would risk to say that there is no such need, but it would
be problematic to alter velocity/density field for solution stability test
and similar ones. In most cases we don't need that.

I could alter unit test to match that current CLB state. Full solution
alteration could be added to TODO list imho

2016-12-14 14:01 GMT+01:00 Łukasz Łaniewski-Wołłk <notifications@github.com>
:

llaniewski:
Please change the test so that it will sum up the "Rho" for instance, ok?

And while you're at it: csv stands for Comma Separated Values, so please write a dat file or write the output/drop_sumF.csv file with commas and header.

PR #118, opened 2016-12-14
Minor spelling fix in python integration

shkodm:
Just fix of minor irritating error in Python integration.


llaniewski:
@shkodm It would be nice if you would do it in v6.1_candidate branch, because we're working on Handlers there and we will overwrite develop with it soon.

llaniewski:
Added to [v6.1 RC] with #122. Closing here, as not vital for maintenance of old version.

Issue #120, opened 2016-12-15
Example refactoring


mdzik:
@TravisMitchell  Pointed out, that examples need to be restructured in the same manner as models.
I'm working on that

mdzik:
Done for now

PR #121, opened 2016-12-15
Move examples

mdzik:
I've moved examples. Basic structure is same as for models.
I hope that I figured out most of them, but A LOT of further clean-up is needed


mdzik:
For #120

llaniewski:
@shkodm, @wojtasMEiL, Please write if your examples are in right folders.

shkodm:
Regarding  a186c67575a3b0d9c0cbbac1f7ac1839707f7f2e , maybe folder "geometry" would more appropriate, that "data" ?  "data" is ambiguous, in my opinion.

mdzik:
maybe input_data or something? It might be something else than geometry (in general). I think sin.csv also should go there

llaniewski:
@mdzik @shkodm data is good. Leave it. If no other comments, I will merge it after tests complete.

PR #123, opened 2016-12-16
[V6.1 RC] Fixes #101 and new NVCC required option

### LINKED ISSUE #101: Extra xml added after config in export is wrong
I found that XML saving in XXX_config_P00_000000.xml is not working right

The xml file looks like this:

``` xml
<?xml version="1.0"?>
<CLBConfig version="2.0" output="output/">
    <!-- this is OK -->
</CLBConfig>
<Run model="d2q9_kuper">
    <Code version="v6.0-beta-529-g31f6a5f" precision="double" cross="GPU" />
</Run>

```

I get this error during post-processing parsing:

```
xml.sax._exceptions.SAXParseException: /XXXXX/run_config_P00_00000000.xml:47:0: junk after document element
```

Which refers to <Run> part. this is wrong: XML states that document should have single root element

```
% xmllint NX1024_prom/id1/output/run_config_P00_00000000.xml                                                                                                                               ~/praca/bench_MP_Laplace
NX1024_prom/id1/output/run_config_P00_00000000.xml:47: parser error : Extra content at the end of the document
<Run model="d2q9_kuper">
^

```

mdzik:
xmlint not is ok with config, and:
```
 +NV_OPT = -D_FORCE_INLINES 
```
fixes compilation issues on newer GCC/NVCC
after
https://github.com/BVLC/caffe/issues/4046


src/Handlers/MainContainer.cpp
```diff
@@ -19,7 +22,8 @@ int MainContainer::Init () {
 			c.append_attribute("cross").set_value("GPU");
 		#endif
 		solver->configfile.save_file(filename);
```
llaniewski: You cannot write the xml file after the execution, because the execution can fail, or can be stopped - and you would still want the xml file to be created.

src/Handlers/vHandler.cpp
```diff
@@ -204,16 +204,17 @@ vHandler * getHandler(pugi::xml_node node) {
 		ret = new acSyntheticTurbulence;
     } else if (name=="Run") {
 		output("Skipping 'Run' element");
-    }else {
+        ret = new Callback();
```
llaniewski: Why generic Callback? I think it should be NULL. If there is segfault because of it - this should be solved, where the segfault occurs.

mdzik: because: https://github.com/CFD-GO/TCLB/blob/v6.1_candidate/src/Handlers.h#L17

It does fault with NULL pointer. I think that returning NULL where pointer-to-object is expected is design error. 

I could change it to 
```C++

		if (hand != NULL) {
     		        hand->solver = solver_;
			int ret = hand->Init();
			if (ret) {
				delete hand;
				hand = NULL;
			}
		}
		ref = new int;
		*ref=1;
debug0("Handler shared pointer: create\n");
```

but other methods of Handler might expect it to be valid object somewhere

src/Handlers/MainContainer.cpp
mdzik: BUT we don't want to process altered XML. 
I might add save of deep copy instead of solver->configfile directly

src/Handlers/vHandler.cpp
llaniewski: Ok, please add a NullHandler then.

src/Handlers/MainContainer.cpp
llaniewski: I see no problem in processing the altered XML. Maybe even in future we will want to add things to the XML on-the-go (by python for instance)

llaniewski:
Closes #101

PR #126, opened 2016-12-19
[V6.1 RC] Adding Factory for Handlers

llaniewski:
Adding a factory pattern for Handlers.

Registering is done by writing a function, which checks the node if it is proper for a specific Handler and creates it when it is. The thing is done with:
```c++
vHandler * Fun(const pugi::xml_node& node) {
  std::string name = node.name();
  if (name == "VTK") {
                return new cbVTK;
  }
  return NULL;
}

template class HandlerFactory::Register< Fun >;
```

I think the syntax is quite transparent. I added all the needed registrations based on the old getHandler function (thank god for grep/sub/regexec and some R magick).

We will finally get rid of this: https://github.com/CFD-GO/TCLB/blob/v6.1_candidate/src/Handlers/vHandler.cpp#L90


llaniewski:
@mdzik @shkodm Where is my review? :D

mdzik:
I think we could avoid that 
```C++
vHandler * Ask_For_GenericContainer(const pugi::xml_node& node) {
+  std::string name = node.name();
+  if (name == "Units") {
+		return new GenericContainer;
+  }
+  return NULL;
+}
```
with some MPL magic. But for now, maybe make macro
```C
#register_me(name,surename)
```
so we wont need to change all of them i the future?

llaniewski:
When I search MPL magic I get this: http://pl.mlp.wikia.com/wiki/My_Little_Pony_Przyja%C5%BA%C5%84_to_magia_Wiki

I think registering a function is quite transparent. And general enough to allow for this:
https://github.com/llaniewski/TCLB/blob/v6.1_candidate/src/Handlers/acUSAdjoint.cpp#L54

We can add a macro - but I (and I'm a fan of macros) think it would obfuscate things.

mdzik:
Ponies are nice, but I was thinking about something like
http://www.boost.org/doc/libs/1_63_0_b1/libs/mpl/doc/index.html
but that could be a really big mess, so it's only a suggestion. 

USAdjoint is a naming mess not a good example :p

The fact that you could break relation
xml node name <-> handler class
is a bad, not a good thing

parameter should influence only a behaviour of the Handler not it's class (with only possible exception of derived classes, but I'm not sure about that). I know, that for now it does not change a think, but could be a problem in the future. XML case file should be similar to serialized objects pipeline.

At the moment you call AskForXXX you already know, that you will get XXX class object, not YYY

mdzik:
I think, something like that should work:
```C++
struct MyXMLName { static const std::string name; }
const std::string MyXMLName::val = "USAdjoint";
XXXX::Register<USAAdjoint, MyXMLName>;
```
and
```C++
       template <Product* (*T)(const Input&), XMLNodeName>
	class Register : public Worker {
	        // Static member, which will register this worker in it's constructor
		typedef RegisterMe< Register< T > > Idiot;
		static Idiot Dummy;
	public:
		virtual Product* Produce(const Input& input) { 
                    if (XMLNodeName::val == input->nodename {
                     return T(input);
                    }
                 };
	};	
```

or we could just add static const field xmlname to asUSAdjoint class. We pass node to init either way so there is no need for additional pre-preparation?

llaniewski:
I think a dynamic dispatch is more interesting (although I agree it is not very xml-compliant). Even when you think about producing the same type of object - you can call different Constructors, based on node attributes etc. Or provide different arguments to the constructors. This would be nice in terms of switching to more C++ coding.

Also when you think about the future, and whole xml setup, it would be useful to create the same object (with different constructor arguments) for a set of different xml nodes (for instance, the names of NodeTypes in Geometry).

For a `<Wall/>` element you want `new Brush(NODE_WALL)`, for `<EVelocity/>` a `new Brush(NODE_EVelocity)`, etc.

I'm still thinking on a good DOM-like structure - but I do not have a working prototype.

If you think it would make things more transparent, I can add a `static std::string xmlname` to handlers, and register some kind of default checker, for the "trivial" handlers, and run the rest with the function mechanism.

@mdzik Please make a review with accept/request changes/etc in github.

@shkodm Please write if you do not have time now for any reviewing work.

shkodm:
@llaniewski , wait, I will go through it and add a review tomorrow or so.

shkodm:
@llaniewski , after 43f8c566c63d3614614613143e3159e6e4e4b68e Catalyst handler would get registered, even if it was disabled in configure. You neglected ifdef's.

llaniewski:
@shkodm Corrected.

PR #131, opened 2017-02-01
[V6.1 RC] Adding Options mechanism

llaniewski:
Added an mechanism to add optional functionality to models. The mechanism is triggered by setting `OPT` variable in the conf.mk file in the model directory. For example:
```bash
OPT="(A+B)*C"
```
Will add three optional functionalities: A, B and C - called Options. The summation means that two options are exclusive (you can have A or B), and multiplication allows for their combination (you can have A, C and both). The mechanism allows to compile a model for all the resulting combinations. The name of the model is created by adding the option after an underscore. In this case:

Model | A | B | C
--- | --- | --- | ---
model | | |
model_A | x | |
model_B | | x |
model_C | | | x
model_A_C | x | | x
model_B_C | | x | x

All these models are compiled from the same Dynamics.R and Dynamics.c, but with different options set to TRUE or FALSE in the Options R variable. One can switch on and off parts of the code with:
```R
if (Options$A) {
# code for option A
} else {
# code without option A
}
```
or in pure C/C++, with:
```c
#ifdef OPTIONS_A
// code for option A
#else
// code without option A
#endif
```


mdzik:
Coool :D
- Do we have MODEL_all compilation?
- Models would be similar, maybe some sort of vti outfiles suffix so results wont mix-up?

llaniewski:
Do we need compilation of model for all combinations at once? I don't see a use-case now.

As for outfiles - for now, as earlier, the information about the model that was executed is only in the `output/..._config_....xml` file. If you think the model name should be added to the output files names, maybe at it as an enchancement issue.

llaniewski:
@shkodm I think this functionality may be of interest for you. I think that later on (after we finally merge everything to develop branch) we should make the d3q27_cumulant averageing as such an option.

shkodm:
@llaniewski, Yes, applying this to averaging in cumulant model was the first thing, that I thought about, upon seeing this pull request.

llaniewski:
@mdzik git?

Issue #144, opened 2017-05-15
Travis-CI support for CUDA


llaniewski:
There is a persisting problem in support for CUDA on updated Travis-CI virtual machines. The update occurred on May 4. Our compilations encountered error first here:

https://travis-ci.org/llaniewski/TCLB/jobs/229029926

Suggested remedies:
- Force improvement of the nvidia apt source - *no idea how to do it*
- Move to container-based infrastructure, as cuda packages are included in their whitelist - *in progress* #145 
- Keep the virtual machine based setup, but use apt: addon for cuda packages - *seems to work* #142

llaniewski:
Solved for now by #142

PR #145, opened 2017-05-15
Move to Travis container infrastructure

llaniewski:
Changed to Travis-CI container infrastructure (no sudo) to go around #144


mdzik:
I would vote to drop it: compile time (d2q9)
with sudo: ~7min
in container ~17min
not worth it

PR #149, opened 2017-08-14
New 2D multiphase model and example file

TravisMitchell:
I added the mode d2q9_pf_velocity. This is a pressure/velocity based LBM modified from the work of Zu and He (2013), coupled with the conservative phase field model of Geier et al. (2015).

I have also included a general test case file to simulate an arbitrary Rayleigh-Taylor instability.
A couple of options are specified in this model if the user feels the need to (A) include higher order terms in the force incorporation or to (B) use the temporal derivative term used by Ren et al. (2016) in their proposed phase field model.

One note of precaution for users is that this model does not yet have three-phase contact conditions implemented, so is currently restricted to cases where this behaviour does not occur.


models/multiphase/d2q9_pf_velocity/conf.mk
```diff
@@ -0,0 +1,3 @@
+ADJOINT=0
+TEST=FALSE
+OPT="A+B"
```
llaniewski: It would be good to have more meaningful names for the options? and a # comment on what is what in this file.

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
```diff
@@ -0,0 +1,714 @@
...
+}
+CudaDeviceFunction real_t getP(){
+	real_t   p = <?R C(sum(g)) ?>;
+	return p;
+}
+
+// 	HELPER FUNCTIONS:
+CudaDeviceFunction vector_t calcGradPhi(){
+	vector_t gradPhi = {0.0,0.0,0.0};
+	if (NodeType & NODE_SymmetricXY_W) {
```
llaniewski: If you have multiple nodetypes in one group it is a bad idea to use "&" for checking. It should be:
```c
if ((NodeType & NODE_ADDITIONALS) == NODE_SymmetricXY_W) {
```

TravisMitchell:
Hi @llaniewski I think it is ready to merge, I had a strange error the other day while compiling with Rinside about one of my variable names though - so I switched the name minorly. I also added an XPipe geometry into the Geometry.cpp.Rt file as I was using this to initialise Taylor bubbles, is this okay to add in?

llaniewski:
@TravisMitchell Can you comment a bit more on the error you get with RInside and this "Width" parameter?

If you will stick with the "IntWidth" then please change it in the example "xml" files.

TravisMitchell:
@llaniewski I think I will stick with IntWidth ("Interface Width"). The error may have come due to the fact I was building over a pre-code generated model, namely hadn't removed the files from CLB/d3q27.... after playing around with a few things? I can't say I really looked into it for long, I had just seen it was getting angry at the variable so did a switch through the name as a preliminary check and it built fine, so I had moved on with what I had originally been doing (sorry).

PR #152, opened 2017-09-30
Changes to d3q27_cumulant

PabloOb:
I modified the model d3q27_cumulant - added options to run with Interpolated Boundaries and Smagorinsky SGS model (mention the original author with @llaniewski @shkodm ). 
I fixed issues with src/Geometry.cpp.Rt which were causing problems with  Interpolated Boundaries

Additionally changed constant in synthetic turblence generator (a=1.453)


src/Geometry.cpp.Rt
```diff
@@ -492,8 +492,8 @@ inline int Geometry::loadSTL(lbRegion reg, pugi::xml_node n)
     ret = fread(header, 80, sizeof(char), f);
     ret = fread(&ntri, 1, sizeof(int), f);
     tri = (STL_tri *) malloc(ntri * sizeof(STL_tri));
-    char *lev = (char *) malloc(reg.size() * sizeof(char));
-    for (int i = 0; i < reg.size(); i++)
+    char *lev = (char *) malloc(reg.sizeL() * sizeof(char));
+    for (size_t i = 0; i < reg.size(); i++)
```
llaniewski: `size()` have to be changed to `sizeL()` to match the type of `i`. 

`size()` returns an `int`, and `sizeL()` returns a `size_t`

models/flow/d3q27_cumulant/Dynamics.R
```diff
@@ -25,44 +26,61 @@ AddSetting(name="GalileanCorrection",default=1.,comment='Galilean correction ter
...
 
 AddDensity( name="SynthTX",dx=0,dy=0,dz=0)
 AddDensity( name="SynthTY",dx=0,dy=0,dz=0)
 AddDensity( name="SynthTZ",dx=0,dy=0,dz=0)
 
+AddNodeType("IB",group="HO_BOUNDARY")
+for (f in fname) AddField(f,dx=0,dy=0,dz=0) # Make f accessible also in present node (not only streamed)
+
 #Averaging values
 if (Ave) {
```
llaniewski: It would be good to make the averaging also an option.

models/flow/d3q27_cumulant/Dynamics.c.Rt
```diff
@@ -142,45 +170,101 @@ CudaDeviceFunction void SSymmetry()
 <?R Symmetry(2,1) ?>
 }
 
-CudaDeviceFunction void Run() {
+
+CudaDeviceFunction void InterpolatedBounceBack()
```
llaniewski: It would be good to also `#ifdef` this function.

llaniewski:
@PabloOb Please correct the things from the review, so we can continue with the merge.

models/flow/d3q27_cumulant/Dynamics.R
PabloOb: I've tried to do that. Unfortunately, encountered some problems along the way 
* Using #ifdef OPT_AVG here resulted in compilation error (similar for all averaged variables): 
Dynamics.c:71:16: error: ‘varUXUY’ was not declared in this scope
         st.z = varUXUY/average_iter - getavgU().x*getavgU().y;
                ^~~~~~~
makefile:48: recipe for target 'cuda.o' failed

* Additionally, I've tried to pass the value of OPT_AVG to Dynamics.R, not to create unnecessary fields if averaging is turned off, but it is not possible.

.idea/libraries/R_User_Library.xml
```diff
@@ -0,0 +1,6 @@
+<component name="libraryTable">
```
llaniewski: This file shouldn't be here

models/flow/d3q27_cumulant/Dynamics.c.Rt
```diff
@@ -27,19 +27,32 @@ CudaDeviceFunction real_t getRho(){
...
-	u.y = (u.y + ForceY/2.)/d;
-	u.z = (u.z + ForceZ/2.)/d;
-	return u;
+	RunBoundaries();
+	return getU_();
+
 }
-#ifdef avgUX
+
+#ifdef OPT_AVG
```
llaniewski: Aha! According to documentation, it should be "OPTIONS_AVG" not "OPT_AVG"

http://cfd-go.github.io/TCLB/model_info.html
https://github.com/CFD-GO/TCLB/pull/131

So, that was the culprit.

models/flow/d3q27_cumulant/Dynamics.R
```diff
@@ -25,44 +26,61 @@ AddSetting(name="GalileanCorrection",default=1.,comment='Galilean correction ter
...
 AddNodeType("SPressure", "BOUNDARY")
+AddNodeType("Buffer", "BOUNDARY")
+
+AddNodeType("NSymmetry", "ADDITIONALS")
+AddNodeType("SSymmetry", "ADDITIONALS")
+AddNodeType("Body", "BODY")
 	
 #Adding terms for supporting time-correlation for synthetic turbulence
 
 AddDensity( name="SynthTX",dx=0,dy=0,dz=0)
```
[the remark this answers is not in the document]
llaniewski: @PabloOb and remember to get rid of SynthTX etc, so we won't get the initialization error (they are not used!)

PR #155, opened 2017-11-10
Fix of d3q27_pf_velocity MRT matrix

TravisMitchell:
I modified model d3q27_pf_velocity to fix the MRT matrix, previously having issues with orthogonality.

Symmetry boundaries were also added to d2q9_pf_velocity.


models/multiphase/d2q9_pf_velocity/Dynamics.R
```diff
@@ -105,3 +105,5 @@ AddGlobal(name="RTISpike",  comment='Spike Tracker')
 # Boundary things
 AddNodeType(name="MovingWall_N", group="BOUNDARY")
 AddNodeType(name="MovingWall_S", group="BOUNDARY")
```
mdzik: SSymmetr NSymmetr etc

llaniewski: @mdzik We will soon move to unified naming for symmetries (and automatic symmetries). This will be corrected in all models.

PR #164, opened 2017-12-07
Fixing bug related to wide access (accessing data in wider regions)

llaniewski:
### Interior/Border
Corrected the division between Interior and Border elements, for access directions with dy/dz values higher then 1

## Wide access all on one side
Fixed a bug, which was causing wrong access conditions if the dx/dy/dz were all eigther positive or negative, and had a spread. E.g. `minx=-2` and `maxx=-1`


llaniewski:
@TravisMitchell fixed. You can pull from upstream now.

PR #170, opened 2018-01-09
Update Dynamics.c.Rt

ggruszczynski:
what is the purpose of the j loop?


llaniewski:
It's a question for @TravisMitchell 
But if I understand the model right, this is a fixed point iteration, for which (in most cases) it is sufficient to do 2 iterations.

ggruszczynski:
Ok, @TravisMitchell said that more numerical tests are needed to figure out how much it helps.

PR #171, opened 2018-02-07
CentOS install script

ggruszczynski:
I updated the install.sh to make it working with CentOS.

I tested the script on VM with:
LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.4.1708 (Core) 
Release:        7.4.1708
Codename:       Core

Unfortunately, I don't have GPU to test, thus cuda part is missing from the script.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -2,7 +2,7 @@
 
 # --------------- UTILITY FUNCTIONS -------------------------
 function usage {
-	echo "install.sh [dry] [skipssl] r|rdep|cuda|submodules|openmpi|coveralls|python-dev|rpython [VERSION]"
+	echo "install.sh [dry] [skipssl] Ubuntu|CentOS [OS] r|rdep|cuda|submodules|openmpi|coveralls|python-dev|rpython|modules [VERSION]"
 	exit -2
 }
 
@@ -26,7 +26,7 @@ function rm_tmp {
   return 0;
 }
   
-# --------------- First check if running dry ----------------
+# --------------- First, check if running dry ----------------
 if test "x$1" == "xdry"
 then
 	RUN=dry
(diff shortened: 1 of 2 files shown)


tools/install.sh
```diff
@@ -171,11 +222,30 @@ submodules)
...
+	;;
+module)
+	try "Downloading module" wget https://github.com/cea-hpc/modules/releases/download/v4.1.0/modules-4.1.0.tar.bz2
+	try "Unpacking archive" tar -xjf modules-4.1.0.tar.bz2 -C .
+	try "Entering module directory" cd modules-4.1.0
+	try "Configuring" ./configure
+	try "make" make
+	try "make install" make install
+	try "Leaving module directory" cd ..
+	try "Restarting terminal" . ~/.bashrc
```
llaniewski: This won't persist in the terminal (all changes of environment are local to the script. Better would be a comment "Remember to restart terminal"

tools/install.sh
```diff
@@ -41,6 +41,20 @@ then
 	shift
 fi
 
+# ------------------- Second, check OS  ----------------------
+OS=""
+if lsb_release -sid | grep "CentOS"
```
llaniewski: I think it's better to be dependent on having a apt or yum command, then on specific os name.

ggruszczynski: I agree, now scripts checks whether yum or apt-get exists.

tools/install.sh
ggruszczynski: Good to know, I updated the comment. 
I left the 'try "Remember to restart terminal" . ~/.bashrc' in case some CI VM would need it one day.

PR #175, opened 2018-03-20
Changes to cuts

PabloOb:
Changed type of cut_t from 'unsigned char' to 'unsigned short int', updated NO_CUT and CUT_MAX in Geometry.cpp.Rt. Rounding near-zero(<1e-10) numbers in GaussSolve in unit.cpp.


PabloOb:
Resolved the problem with cuts in d3q27_cumulant, by changing constant value to 2.0/CUT_MAX

PR #181, opened 2018-07-30
gcc 7 compilation errors

mdzik:
fix note: ‘__builtin___sprintf_chk’ output between X and X bytes into…a destination of size 1024 warnings in gcc7

```
The -Wformat-overflow=level option detects certain and likely buffer overflow in calls to the sprintf family of formatted output functions. Although the option is enabled even without optimization it works best with -O2 and higher.

For example, in the following snippet the call to sprintf is diagnosed because even though its output has been constrained using the modulo operation it could result in as many as three bytes if mday were negative. The solution is to either allocate a larger buffer or make sure the argument is not negative, for example by changing mday's type to unsigned or by making the type of the second operand of the modulo expression unsigned: 100U.

void* f (int mday)
{
  char *buf = malloc (3);
  sprintf (buf, "%02i", mday % 100);
  return buf;
}

warning: 'sprintf may write a terminating nul past the end of the destination [-Wformat-overflow=]
note: 'sprintf' output between 3 and 4 bytes into a destination of size 3
```
see: https://gcc.gnu.org/gcc-7/changes.html

### DIFF
--- a/src/Handlers/MainContainer.cpp
+++ b/src/Handlers/MainContainer.cpp
@@ -4,7 +4,7 @@ std::string MainContainer::xmlname = "CLBConfig";
 
 int MainContainer::Init () {
 		GenericAction::Init();
-		char filename[STRING_LEN];
+		char filename[2*STRING_LEN];
 
 		solver->outIterFile("config", ".xml", filename);
 		pugi::xml_node n = solver->configfile.child("CLBConfig").append_child("Run");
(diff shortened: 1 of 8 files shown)


mdzik:
This is also an indication of potential overflow: we used buffer of size STRING_LEN everywhere. Even when we put together  other bufffer of same size + something else (like in filenames)

PR #182, opened 2018-08-03
Minor additions to TM phase field models

TravisMitchell:
Only small adjustments, but so there is not any large divergence.

I modified the model:
   - d2q9_SRT: student couldn't run d2q9_srt_example.xml, adjustments so that this now runs okay.
   - d2q9_pf_velocity: play with convective outlet
   - d3q27_pf_velocity: 
        - Added smoothing section for assisting initialisation from recommendation of gg
        - Convective and velocity boundaries
        - Three phase contact
        - Some initialisation geometries that I really need to write into geometry rather than here.

### DIFF
--- a/models/flow/d2q9_SRT/Dynamics.c
+++ b/models/flow/d2q9_SRT/Dynamics.c
@@ -53,12 +53,11 @@ CudaDeviceFunction void Run() {
 //		NVelocity();
 //		break;
     }
-	if (NodeType & NODE_MRT) 
+	if (NodeType & NODE_BGK) 
 	{
-	// Set as if MRT as majority of examples specify
-	// solution zone as MRT box, so avoid changing 
-	// input files.
 		CollisionBGK();
+	} else if (NodeType & NODE_MRT) {
+		printf("NOT IMPLEMENTED... press <crtl + c>\n");
 	}
 }
 
--- a/models/flow/d2q9_SRT/Dynamics.c
+++ b/models/flow/d2q9_SRT/Dynamics.c
@@ -56,9 +56,7 @@ CudaDeviceFunction void Run() {
 	if (NodeType & NODE_BGK) 
 	{
 		CollisionBGK();
-	} else if (NodeType & NODE_MRT) {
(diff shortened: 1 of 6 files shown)


models/flow/d2q9_SRT/Dynamics.c
```diff
@@ -53,12 +53,11 @@ CudaDeviceFunction void Run() {
...
     }
-	if (NodeType & NODE_MRT) 
+	if (NodeType & NODE_BGK) 
 	{
-	// Set as if MRT as majority of examples specify
-	// solution zone as MRT box, so avoid changing 
-	// input files.
 		CollisionBGK();
+	} else if (NodeType & NODE_MRT) {
+		printf("NOT IMPLEMENTED... press <crtl + c>\n");
```
mdzik: Is this CUDA/GPU safe?

models/multiphase/d2q9_pf_velocity/Dynamics.R
```diff
@@ -39,9 +39,11 @@ AddField('PhaseF',stencil2d=1, group="PF")
 #	Additional access required for outflow boundaries
 if (Options$Outflow){
 	for (d in rows(DensityAll)){
-    		AddField( name=d$name,  dx=-d$dx-1, dy=-d$dy, dz=-d$dz )
+    		AddField( name=d$name,  dx=-d$dx-1, dy=-d$dy )
+    		AddField( name=d$name,  dx=-d$dx, dy=-d$dy-1 )
```
mdzik: same name?

llaniewski: @mdzik This extends the scope avaliable for reading for field.
@TravisMitchell You can also do `AddField( ..., dx=c(-d$dx-1,-d$dx), ...)`

llaniewski:
@TravisMitchell Let's merge TravisMitchell/TCLB#1 before we continue.

mdzik:
seems ok. those models should have some tests...
I think the way that test are separated in sub/module makes in even more difficult to maintain (fork point to "wrong" repo)
I think other way/some script is needed. Just saying ;)

PR #184, opened 2018-08-14
Fix debug build in gcc 7 

mdzik:
do not add int to void

### DIFF
--- a/src/LatticeContainer.inc.cpp.Rt
+++ b/src/LatticeContainer.inc.cpp.Rt
@@ -17,7 +17,7 @@
 #include "Lattice.h"
 #include "mpi.h"
 #define ALLOCPRINT1 debug2("Allocating: %ld b\n", size)
-#define ALLOCPRINT2 debug1("got address: (%p - %p)\n", tmp, tmp+size)
+#define ALLOCPRINT2 debug1("got address: (%p - %p)\n", tmp, (long)tmp+(long)size)
 
 int _xsdim = 0;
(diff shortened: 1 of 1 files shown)


mdzik:
No its not OK with C standard & pedantic mode

https://stackoverflow.com/questions/3523145/pointer-arithmetic-for-void-pointer-in-c

```
LatticeContainer.inc.cpp: In member function ‘void LatticeContainer::ActivateCuts()’:
LatticeContainer.inc.cpp:17:64: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
 #define ALLOCPRINT2 debug1("got address: (%p - %p)\n", tmp, tmp+size)
                                                             ~~~^~
Global.h:170:23: note: in definition of macro ‘DEBUG1’
     #define DEBUG1(x) x; fflush(stdout);
                       ^
LatticeContainer.inc.cpp:17:21: note: in expansion of macro ‘debug1’
 #define ALLOCPRINT2 debug1("got address: (%p - %p)\n", tmp, tmp+size)
                     ^~~~~~
LatticeContainer.inc.cpp:477:17: note: in expansion of macro ‘ALLOCPRINT2’
                 ALLOCPRINT2;


```

llaniewski:
Ok. Let's go then with (unsigned char*) in stead of (void*). So something like:
```c
#define ALLOCPRINT2 debug1("got address: (%p - %p)\n", (unsigned char*)tmp, (unsigned char*)tmp+size)
```

PR #194, opened 2019-01-22
D2q9 pf cm

ggruszczynski:
Cascaded (CLBM) collision scheme is introduced for both hydrodynamics and phase-field.
Based on the d2q9_pf_velocity implemented by @TravisMitchell


models/multiphase/d2q9_pf_cm/Dynamics.c.Rt
```diff
@@ -0,0 +1,1488 @@
...
+	f_in[2] = 0.5*temp[2] + 0.5*temp[4] - 0.5*temp[6] - 0.5*temp[8];
+	f_in[3] = -0.5*temp[1] + 0.5*temp[3] + 0.5*temp[7] - 0.5*temp[8];
+	f_in[4] = -0.5*temp[2] + 0.5*temp[4] + 0.5*temp[6] - 0.5*temp[8];
+	f_in[5] = 0.25*temp[5] + 0.25*temp[6] + 0.25*temp[7] + 0.25*temp[8];
+	f_in[6] = -0.25*temp[5] + 0.25*temp[6] - 0.25*temp[7] + 0.25*temp[8];
+	f_in[7] = 0.25*temp[5] - 0.25*temp[6] - 0.25*temp[7] + 0.25*temp[8];
+	f_in[8] = -0.25*temp[5] - 0.25*temp[6] + 0.25*temp[7] + 0.25*temp[8];
+}
+
+CudaDeviceFunction void relax_and_collide_CM_phase_field(real_t f_in[9], real_t tau, vector_t u, vector_t F_phi, vector_t Fhydro, real_t rho)
```
ggruszczynski: `vector_t Fhydro` is useless here

TravisMitchell: @ggruszczynski

Any opinion on the above?

ggruszczynski: Naming:
- I will keep the `d2q9_pf_velocity` and override the old model.
- old mrt model was also in 'in press' but was placed in models/multiphase not models/article folder. I think that physics-related structure is clearer.

BC: Yes, I will add a switch.

llaniewski:
@TravisMitchell Are we happy? Can we merge?

TravisMitchell:
Yes, I think we are okay to merge. But @ggruszczynski, if you are looking to do a similar collision scheme for the 3D model - please build this in as a new collision scheme into the already tested, verified and validated d3q27_pf_velocity model. I actually find the d2q9 model you have put together for this commit to be slower than my original d2q9_pf_velocity model as there seems to be quite a large number of sub-functions and additional code that could be neatened up with the use of the RT template.

ggruszczynski:
Thx for review @TravisMitchell . I prefer to split functionalities to functions even at the price of slower performance. I understand that the d3q27_pf_velocity has been rigorously tested thus I won't refactor working code.

llaniewski:
Oh, @ggruszczynski ,  you should not have had the access to merge.
Sorry, but I have to hard reset this mess.

ggruszczynski:
;(

PR #202, opened 2019-02-18
refreshed wave example + new ADE example

ggruszczynski:
PDE: I refreshed the wave2D model and added new ADE model. Both with examples.


llaniewski:
@ggruszczynski if `wave2D_obsolete` is obsolete, why not delete it?

ggruszczynski:
@llaniewski There is some adjoint stuff, thus I am not sure. 
Let me know if you need it or not.

llaniewski:
@ggruszczynski If somebody will want to see it, it's still in the git history. Delete it.

PR #208, opened 2019-03-22
Requesting package manager only when needed in tools/install.sh

llaniewski:
`tools/install.sh` was discovering the package manager even when it didn't need it. I moved it into a function, which is called only when needed.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -57,7 +57,6 @@ function get_PMS {
 		if test -z "$PMS"
 		then
 			echo "Unknown type of Package Manager, only apt-get and yum are supported."
-			usage
 			exit 2;
 		fi
 	fi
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -43,17 +43,25 @@ fi
...
+			if [ -x "$(command -v $i)" ] ; then 
+				echo "Discovered Package Manager: $i"
+				PMS=$i
+			fi
+		done
+		if test -z "$PMS"
+		then
+			echo "Unknown type of Package Manager, only apt-get and yum are supported."
+			usage
+			exit 2;
```
ggruszczynski: `exit 2;` is useless, because the `usage` function makes `exit -2`

PR #222, opened 2019-09-25
d2q9 immiscible 2-component model with some neat features 

mdzik:
I added/upgraded  the model d2q_csf with numbers of submodels. This is subject of pending publication - references will be added when avealible.
This is only draft PR - comments&questions etc. welcome  

Pending PR #221 
TODO: remove some added features to separate commits


models/multiphase/d2q9_pf/Dynamics.R
```diff
@@ -27,36 +27,37 @@ AddDensity( name="h[6]", dx=-1, dy= 1, group="h")
...
-
-
-
-AddSetting(name="OverwriteVelocityField", default="0")
-#########################################################
-
-
-# THIS QUANTITIES ARE NEEDED FOR PYTHON INTEGRATION EXAMPLE
+# THIS DENSITIES ARE NEEDED FOR PYTHON INTEGRATION EXAMPLE
 # COMMENT OUT FOR PERFORMANCE
```
llaniewski: Can we have a `BC` option here also?

mdzik: I've added no_bc instead  - tests and examples rely on the presence of bc. This is a demonstrator for "advection" model, so injecting velocity filed makes a lot of sense here

PR #230, opened 2019-10-18
Some fixes for Geometry and optional C++11 support

ggruszczynski:
Adding Cylinder to Geometry.
Fixes STL cuts export, which crashed ParaView.
Added optional C++11 support.


.travis.yml
```diff
@@ -57,11 +58,11 @@ install:
 
 script:
     - make configure
-    - ./configure --enable-double --with-cuda-arch=sm_20 --enable-marklines --with-python --enable-paranoid
+    - ./configure --enable-double --with-cuda-arch=sm_30 --enable-marklines --with-python --enable-paranoid
```
ggruszczynski: it seems that CudaThreadSynchronize has been deprecated and replaced by CudaDeviceSynchronize in sm_30
thus --enable-paranoid catches the depreciated warning and crashes.

llaniewski:
@ggruszczynski make a proper description, which explains what you did. It's a mix of many things, and it does not compile on Travis-CI (I guess it's the C++11 part)

src/conf.R
```diff
@@ -287,6 +287,7 @@ AddDescription = function(short, long) {
 
 AddNodeType("BGK","COLLISION")
 AddNodeType("MRT","COLLISION")
+AddNodeType("CM","COLLISION")
```
llaniewski: W would not want this by default. In fact as we go I would not want any NodeTypes by default, and all defined in models.

src/configure.ac
```diff
@@ -527,6 +531,14 @@ then
 	NVFLAGS="${NVFLAGS} --ptxas-options=-v"
 fi
 
+
+if test "x${enable_cpp11}" != "xno"
```
llaniewski: Don't make it by default. For now it will make many problems on several architectures.

ggruszczynski:
I will back to this PR later.

Issue #132, opened 2017-02-07
outfile marking to prevent overwrite/mixup


mdzik:
@mdzik :
Models (with variants) would be similar, maybe some sort of vti outfiles suffix so results wont mix-up?

@llaniewski :
As for outfiles - for now, as earlier, the information about the model that was executed is only in the output/..._config_....xml file. If you think the model name should be added to the output files names, maybe at it as an enchancement issue.

options ? :
- tag keyword in xml, that would add tag to ALL possible output files
- do-not-overwrite-a-thing - if outfile exist use date-base prefix/tag

PR #235, opened 2019-10-25
XPath based modification

llaniewski:
# Description
I added support for adding small (and big) modifications to the xml which is being run on the command line.

## How it works
The syntax quite simple. For instance if you want to set the `output` attribute of your case, that would be:
```bash
CLB/d2q9/main file.xml @output = "test/"
```

Where `@output` can be any [XPath](https://en.wikipedia.org/wiki/XPath) expression matching some attributes. For instance:
```bash
CLB/d2q9/main file.xml //@output = "test/"
```
Would substitute `output` attribute with `test/` *everywhere* in the case file. With this you can do:
```bash
CLB/d2q9/main file.xml //@Viscosity = "0.16"
```
To run your case with a different Viscosity

## Notice
**The xml file (with all modification) is saved in the output directory with all the results for later inspection**

## Operators
I implemented three operators: `=`, `insert` and `print`. One can add new operators later on.

### `print` operator
Just prints what the XPath maches:
#### Examples
```bash
# print all Params attributes:
CLB/d2q9/main file.xml //Params/@* print
```

### `=` operator
Substitutes the matching attributes with `[xpath expression] = [value]` or add attribute to nodes with `[xpath expression] @[attribute name] = [value]`.
#### Examples
```bash
# change output path everywhere in the case file
CLB/d2q9/main file.xml //@output = "test/"
# change viscosity
CLB/d2q9/main file.xml //@Viscosity = "test/"
# change the length of domain
CLB/d2q9/main file.xml Geometry/@nx = "128"
# select field to export in VTK
CLB/d2q9/main file.xml //VTK/@what = "U,Rho"
# select field to export in VTK (even if the attribute didn't exist)
CLB/d2q9/main file.xml //VTK @what = "U,Rho"
```

### `insert` operator
Inserts an arbitrary piece of XML code into the case. You can specify where it should be placed with:

specifier | placement
--- | ---
`after` | after the selected node (sibling)
`before` | before the selected node (sibling)
`last` (default) | at the end of the selected node (child)
`first` | at the beginning of the selected node (child)

#### Examples
```bash
# run some more iterations at the end
CLB/d2q9/main file.xml . insert '<Solve Iterations="100"/>'
# add some parameters
CLB/d2q9/main file.xml Model insert '<Params Viscosity="10"/>'
# add a Log output
CLB/d2q9/main file.xml Model insert after '<Log Iterations="10"/>'
# add VTK output before every Solve
CLB/d2q9/main file.xml //Solve insert before '<VTK/>'
```

### `delete` operator
Deletes what the XPath maches.
#### Examples
```bash
# Delete all VTK output:
CLB/d2q9/main file.xml //VTK delete
```

### DIFF
--- a/src/config.mk.in
+++ b/src/config.mk.in
@@ -16,7 +16,7 @@ SOURCE_CU=Global.cu Lattice.cu main.cu vtkLattice.cu vtkOutput.cu cross.cu cuda.
 SOURCE=$(SOURCE_CU)
 HEADERS=Global.h gpu_anim.h LatticeContainer.h Lattice.h Region.h vtkLattice.h vtkOutput.h cross.h gl_helper.h Dynamics.h Dynamics.hp types.h pugixml.hpp pugiconfig.hpp
 
-OBJ  = vtkOutput.o cuda.o Global.o Lattice.o vtkLattice.o cross.o pugixml.o Geometry.o def.o unit.o Solver.o SyntheticTurbulence.o Sampler.o ZoneSettings.o RemoteForceInterface.o
+OBJ  = vtkOutput.o cuda.o Global.o Lattice.o vtkLattice.o cross.o pugixml.o Geometry.o def.o unit.o Solver.o SyntheticTurbulence.o Sampler.o ZoneSettings.o RemoteForceInterface.o xpath_modification.o
 
 AOUT=main
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -282,6 +282,61 @@ int main ( int argc, char * argv[] )
 		error("Error while parsing %s: %s\n", filename, result.description());
 		return -1;
 	}
+	
+	if (argc > 2) {
+		for (int i = 2; i < argc; i++) {
+			try {
+				output("XPATH: %s\n",argv[i]);
+				pugi::xpath_node_set found = solver->configfile.select_nodes(argv[i]);
+				output("XPATH: %ld things found\n", found.size());
+				i++;
+				if (i < argc) {
+					if (strcmp(argv[i], "=") == 0) {
+						i++;
+						if (i < argc) {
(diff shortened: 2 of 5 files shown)


llaniewski:
@mdzik @TravisMitchell @ggruszczynski Any thoughts?

llaniewski:
You can do even more complex things.

For an xml like:
```xml
...
 <Units>
  <Params D="0.01m" gauge="32"/>
  <Params Viscosity="1e-5m2/s" gauge="0.01"/>
 </Units>
...
```
one can modify the mesh density with:
```xml
CLB/d2q9/main file.xml Units/Params[@D]/@gauge = 64
```

TravisMitchell:
I quite like the idea, it will be useful for grid convergence (or scaling) tests. Could potentially use it for parametric studies as well. Although would need to be slightly careful and be sure to rename the outputs as well each time.

llaniewski:
> [...] Although would need to be slightly careful and be sure to rename the outputs as well each time. (@TravisMitchell )

Yes, this is related to #132 . But I don't really have a nice automatic way to generate the new name, for the output not to be overwritten.

We can opt for **never overwrite** strategy (generating new suffix, or directory every time the code is run), but this can be a bit tricky and I don't know if it's the way we want to go. We can move the discussion on that to #132

As for this PR, one can combine modification of parameters with modification of output directory:
```xml
CLB/d2q9/main case.xml //@Viscosity = 0.01 //@output = "output/low_visc/"
```

TravisMitchell:
A possible addition to the capability of this could be to alter (i.e. find and replace) for node types. I came across this while playing around with the phase field velocity model that has both MRT and CM node types. For which having an inline switch between would have saved having two run files. From discussions with @llaniewski this could also be a useful feature for switching boundary conditions between runs.

Issue #100, opened 2016-08-10
<Params> multiple arguments


mdzik:
The xml file looks like this:

``` xml
    <Params thing="A"  thing-zone="B"/>
```

Where it would work different to

``` xml
    <Params thing-zone="A"  thing="B"/>
```

 because order of attributes matters. BUT it's non conforming with XML standard\* .  I know that it's a bug in xml case file, but it's all over examples etc. and we do support that.
Therefore only supported version should be

``` xml
    <Params thing="B"/>
    <Params thing-zone="A" />
```

because order of child nodes is meaningful in XML.

We could add special name/zone argument/node:

``` xml
    <Params    thing="B"       other-thing="B"  />
    <Params   zone="zone"  thing="B"  other-thing="B" />
```

or

``` xml

       <Params   thing="B"   other-thing="B" >
       <Zonal   name="zone"     thing="B"      other-thing="B" />
       <Zonal   name="zone2"  thing="B"   other-thing="B" />
       </Params>
      <Params yetanother="1"/>
```

or maybe another.

In general this is bug IMHO, and need to be clarified before our miracle DOM-based handling ;)

@llaniewski , @shkodm  I could work on that, but in some cases it will break cases. We could make "converter" but still. And which option is most usefull/most expandable.

Other thing is that XML guys prefer

``` xml
<param>
<name>name</name>
<value>value</value>
</param>
```

or  (I would vote for this, still short but allows <[[CDATA and binary/base64 input easly)

``` xml
<param name="name">value</param>
```

over

``` xml
<param name="value" />
```

*) section 3.1 of the XML recommendation. It says, "Note that the order of attribute specifications in a start-tag or empty-element tag is not significant."',  http://stackoverflow.com/questions/726395/order-of-xml-attributes-after-dom-processing

llaniewski:
Fixed by #241

PR #241, opened 2019-11-05
Migration from Params to Param

llaniewski:
# Description

## Problem

(Issue #100) The `<Params/>` element does not conform to the XML standard practices. It's a legacy syntax, which was simpler, but causes more and more problems

## Solution

I introduced a new element `<Param/>` which can be used for a **single setting only**. Further more it is invoked `<Param name="..." zone="..." value="..."/>`.

### Example
The old settings declared as:
```xml
<Params Viscosity="0.16" Velocity-Inlet="0.01" />
```
now become:
```xml
<Param name="Viscosity" value="0.16"/>
<Param name="Velocity" zone="Inlet" value="0.01"/>
```

# Notes:
## Backwards compatibility
To not disturb backwards compatibility too much, the code parses the XML config file first, and *transforms all `<Params/>` elements to proper `<Param/>` elements*. Additionally it sets a "permissive" flag on the case, to be compatible with old treatment of unknown settings. This means that **old case files should work without modification**.

## Transformation of old files
If you have a xml file and change it to the new method of declaring settings, you can do it from command-line:
```bash
make -j d2q9
CLB/d2q9/main file.xml -s out.xml -x # transforms the Params elements
```
The `-s` option saves the transformed xml, and `-x` exists without running the simulation. **Note:** any model can be used - this is just a XML transform.

## Treatment of unknown settings
Until now, TCLB was ignoring any attributes of `<Params/>` for unknown settings. This is sometimes nice, as one wants to have the same case file for several models (and provide settings for all of them). But that also makes it easy to run a simulation with the wrong settings by misspelling the setting name. For instance `<Params Veloooocity="1"/>` would be just ignored and would cause the `Velocity` setting not to be set.

I introduced the `permissive` attribute, which can be set on any element, which will make TCLB ignore unknown settings. Otherwise TCLB will **error on unknown setting**.

### Example
To supply settings for different models in the same xml file (unknown will be ignored), one can use:
```xml
<Model permissive="true">
<Param name="Velocity" value="1"/>
<Param name="Temperature" value="100"/>
</Model>
```
Unknown settings are not permitted by default. This means that eg. `<Param name="Veloooocity" value="1"/>` will fail, if `permissive` attribute is not set anywhere.

## Units
In the new version, it is discouraged to use false names in the `<Units/>` section. If you want to introduce a scale without setting the value of a specific setting, you should write `<Param value="..." gauge="..."/>` without `name="..."`.

### Example
A set of units:
```xml
<Units>
<Params D="1m" gauge="32"/>
<Params Velocity="1m/s" gauge="0.01"/>
</Units>
```
should now be written as:
```xml
<Units>
<Param value="1m" gauge="32"/>
<Param name="Velocity" value="1m/s" gauge="0.01"/>
</Units>
```

### commandline arguments
I introduced `-s` option for saving and `-x` for exiting. Additionally the code now parses the comments in the XML files, and discards them only after all transforms etc are done. This was needed to preserve them in the transform.
### examples and tests
I transformed all XMLs in `example` and `tests` to the new version. **I didn't check if the examples work**. I deleted the `permissive` attribute from case files in `tests`, for them to check the settings names.

FYI: @mdzik @ggruszczynski @TravisMitchell @bhill23 @ndivaira - this will be part of V6.3


mdzik:
And from now on:
```xml
<a>
   <b>Order of</b>
   <b>parameters are respected</b>
</a>
```
:+1:

Issue #252, opened 2019-12-19
Issue with d2q9_pf_velocity model


TravisMitchell:
Found an issue with d2q9_pf_velocity model. This seems to have been introduced in the CM update version of the code as the behaviour is correctly captured in the original d2q9_pf_velocity model. This was noticed when I was looking to replicate some current work in Yos et al. (2019) for a multiphase and or rtemplate tutorial for the documentation.

The below image shows the behaviour of the d2q9_pf_pressureEvolution and original d2q9_pf_velocity model in comparison to the current d2q9_pf_velocity model (MRT relaxation is used here, CM relaxation goes unstable relatively early):
![Comparison of behaviour](https://user-images.githubusercontent.com/17609184/71139587-4cda5700-225a-11ea-8521-311a7f5c1d9f.png)

The xml run files for the pressureEvolution, original d2q9, as well as current d2q9 (both mrt and cm) can be seen below:

[runFiles.zip](https://github.com/CFD-GO/TCLB/files/3981475/runFiles.zip)

Will look to fix this today, but just a notification for anyone using this model in its current form.

TravisMitchell:
@llaniewski 

On my branches:
- v6.1_candidate-ZuHe_Modified_2D    model: d2q9_pf_velocityCheck
- d2q9_pf_velocity_Fix                           model: d2q9_pf_ZuHe_modified

These are the same models, but are finding different results between the code versions. I have pushed the run file that I am using in with the models

llaniewski:
Update: the difference between versions proved to be the difference in interpretation of the cell position made here: 03032b1db60df352d7099882a570be60f98b2d02 (Dec 14, 2017)

llaniewski:
@TravisMitchell what's the status on this?

TravisMitchell:
@llaniewski the difference between versions was resolved as resulting from when we changed X,Y,Z values in Dynamics.c to reference +0.5 values. I believe this test case brought up a strong dependence on initial conditions with this model that I have not observed elsewhere.

llaniewski:
@TravisMitchell yeah. I was thinking about the issue that you were getting different results/blowup if you centered you case on element, and not between elements. Does it still happen?

TravisMitchell:
I will check now-ish, give me 30 mins

PR #253, opened 2019-12-20
+ Issue with d2q9_pf_velocity clean up and fix + example

TravisMitchell:
+ General tidy up of the model and addition of options feature to allow for various collision operators without any extra code compilation. 

+ Added fixedIterator, needs to be greater than 1 for dynamic cases particularly if continuous phase is gas

+ Double checked grad and laplace operators and chemical potential calculations

+ Added droplet falling example that was where issues were noted.

### DIFF
--- a/example/multiphase/d2q9_pf_velocity/Droplet_Falling_2D.xml
+++ b/example/multiphase/d2q9_pf_velocity/Droplet_Falling_2D.xml
@@ -1,0 +1,52 @@
+
+<?xml version="1.0"?>
+<CLBConfig version="2.0" output="output/">
+<!-- 
+	Model:	d2q9_pf_velocity 
+		Can also be run with d2q9_pf_velocity_CM if the collision type is changed.
+
+	This is the test case from Yos et al. (2019) - Journal of Computational Physics
+	Density ratio:10
+	Viscosity ratio: 6.324555
+	Eotvos number = 100
+	Reynolds = 10
+	t* = 3750
+
+   Created: 	14-08-2017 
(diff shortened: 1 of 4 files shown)


llaniewski:
@TravisMitchell does this fix the main issue in #252? Is the new model in agreement with your old published one?

ggruszczynski:
@TravisMitchell @llaniewski  what are the conclusions regarding #252  ?

TravisMitchell:
@ggruszczynski the result of the updated code now matches the original one that I wrote, and stably simulates the droplet falling test case that I was looking at with MRT and CM. The result is close, but does not match exactly with the pressure evolution model from Fakhari, Lee and Geier that I coded up. However, this seems to be from the model itself rather than alterations made to the code, as such I pushed the changes made above. We also found a bit of strange behaviour based on the initial location of the droplet, which may be an artefact of the central isotropic difference - but this requires further examination.
Current-status

llaniewski:
@ggruszczynski @TravisMitchell I understand this matches now the previously published model and I'm merging this PR.

ggruszczynski:
@TravisMitchell @llaniewski 
As you indicated, the `fixedIterator=2` has an influence on the results what was not noticed in the benchmarks used in the publication:
```
The equation for Fµ is implicit, because the N matrix depends on u, which depends on F. How-
ever, it has been found that there is no need to apply a sub-iteration routine to determine the
velocity more precisely because it does not affect the results. 
```

ggruszczynski:
I have checked the previous release against new (merged) PR, the results are OK :)

PR #261, opened 2020-01-16
Cleanup of tools

llaniewski:
# Description
## Problem
There was a lot of unused or unnecessary tools in the code. They cause compatibility issues on different architectures.

## Solution
Deleted all the unused and most of the unnecessary tools, and cleared them from other files.

# Notes
## Additional
Removing `shared_ptr` from `compare.cpp` for #262 (we don't want C++11 for now)


llaniewski:
fixes #262

PR #263, opened 2020-01-17
Fix logging of RayleighTaylor points in d2q9_pf_velocity

TravisMitchell:
I added an RTemplate file which will be used for an RTI tutorial being added to docs.

Also fixed the logging of Rayleigh-Taylor tracking points in the d2q9_pf_velocity model.


example/multiphase/d2q9_pf_velocity/RTI_example.xml
```diff
@@ -0,0 +1,80 @@
+<?R
```
llaniewski: Please name all RTemplate files with the `.Rt` suffix: `RTI_example.xml.Rt`

PR #265, opened 2020-01-18
Partial support for compilation on Windows with MS Build Tools

llaniewski:
# Description
## Problem
The only way to compile TCLB on windows currently is through Windows Subsystem for Linux, which doesn't include CUDA support.

## Solution
I added a simple route through `cmake` for compilation with MS Build Tools.

## Notes
This is not full on support for first-time users, but a clear starting point for experianced users to compile TCLB win CUDA on windows.


llaniewski:
@TravisMitchell I would appreciate if you would test this. Basic instructions are in the README.

TravisMitchell:
Code up and running, couple of minor adjustments:

- Windows configure file will need to mkdir CLB before config.h and config.mk
- Possibly worth highlighting the export locations:
export PATH=$PATH:/c/Program\ Files\ \(x86\)/Microsoft\ Visual\ Studio/2019/BuildTools/MSBuild/Current/Bin/
export PATH=$PATH:/c/Program\ Files\ \(x86\)/Microsoft\ Visual\ Studio/2019/BuildTools/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/bin/
To avoid confusion with the Programs Files/MSBuild directory that the installer makes.
- Additional note on MPI install to grab the developers SDK as well as exe
- Note about running simulations with the windows executable placed in CLB/<model>/debug

TravisMitchell:
After closer inspection of this, it doesn't seem to be stable/perform in the same way the code does on linux for all models. For example, running the Rayleigh-Taylor case that I recently wrote up for the multiphase tutorial on the website has issues on the windows build.

llaniewski:
As I don't yet know why there is a difference in results on windows for the model reported by @TravisMitchell , I'm moving it to V6.4

PR #266, opened 2020-01-23
Fixing C++11 issues in compare

llaniewski:


llaniewski:
@TravisMitchell This should fix finally problems with C++11 in compare.cpp

PR #271, opened 2020-02-20
Fixing issue with model options

llaniewski:
# Problem
## Description
There is no way now to make a model with options, which cannot be compiled without any option selected.
## Background
The option matrix generation is based on R *formula* syntax. In this syntax one can delete the *intercept* term in the formula by adding `-1` at the end of formula. But it was not supported by our compilation mechanism uptil now.

# Solution
I added the support for intercept term.

## Example
`OPT` | Models generated
--- | ---
`(A+B)*C` | `model`, `model_A`, `model_B`, `model_C`, `model_A_C`, `model_B_C`
`(A+B)*C-1` | `model_A`, `model_B`, `model_C`, `model_A_C`, `model_B_C`
`(A+B)*C-C-1` | `model_A`, `model_B`, `model_A_C`, `model_B_C`

### DIFF
--- a/src/makefile.main.Rt
+++ b/src/makefile.main.Rt
@@ -91,7 +91,7 @@ travis : .travis.yml
 	by(Models, Models$group, function(m) {
 		if (length(m$opts[[1]]) > 0) {
 			mat = do.call(rbind,m$opts)
-			mat = ifelse(mat==1, "X","-")
+			mat = ifelse(mat!=0, "X","-")
 			mat = as.data.frame(mat)
 			row.names(mat) = m$name
 		} else {
--- a/src/models.R
+++ b/src/models.R
@@ -44,11 +44,13 @@ get.models = function() {
 			opts = NULL
 		}
 		if (class(opts) == "formula") {
-			opts = terms(opts)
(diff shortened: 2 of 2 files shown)


mdzik:
So every "add" have implicity +1?, and we need to subtract them?
```
(A+B)*C -> 1 + (A+B+1)*(C+1) -> 1 + C + A + B + AC + BC
```
is this ok? (it's bit mor readable, I think)
```
(A+B-1)*C -> 1 + (A+B-1+1)*(C+1) -> A + B + AC + BC
```

czw., 20 lut 2020 o 05:00 Łukasz Łaniewski-Wołłk <notifications@github.com>
napisał(a):

llaniewski:
Yes. you can say that there is an implicit `1` in all the formulas. This is good, as if one says `A*B` he want the model to have *optional* A and B features (model, model_A, model_B, model_A_B). And that was the behavior already. This PR fixes the case in somebody *explicitly subtracts the 1*. So now `A*B-1` gives the right result.

**notice**: the `-` doesn't work in brackets. So: One cannot do `(A-1)*B`, but should do `A*B-B`

PR #275, opened 2020-04-05
Adding XML Schemas

llaniewski:
# Problem
## Description
There were no verification and auto-completion capabilities for the config xml-s for TCLB

## Details
A standard way for describing the *constraints* on an XML is with XML Schema (.xsd). An XML Schema described the possible elements, attribute names and values. This allows for *off-line* formal verification of an XML, and with the use of software like Eclipse of VS Code, it can be used for autocompletion

# Solution
## Generated XML Schema
I added a RTemplate file which generates a XML schema **for a specific model**. This is needed per model, as the names of the settings, globals, etc. is model specific. This allows for both verification and auto-completion of e.g. setting names.

## XML Catalog
Additionally a XML Catalog is generated which can be used to match namespaces to XML Schema. This means that you can simply add the namespace to the xml:
```xml
<CLBConfig version="2.0" output="output/" xmlns="urn:tclb:d2q9">
```

## Verification
Peek 2020-04-05 14-45

## Auto-completion
Peek 2020-04-05 14-39

# Notes
**Important**: Not all arguments and features are supported yet in the schema. The schema is generated from [doc/elements.yaml](https://github.com/CFD-GO/TCLB/blob/develop/doc/elements.yaml) file, which is in many places lacking.


llaniewski:
I read a bit on the [URN](https://en.wikipedia.org/wiki/Uniform_Resource_Name), and I don't know if it was a good idea to use it. Maybe it's better to use standard URL: `https://tclb.io/model/d2q9` or something similar.

PR #283, opened 2020-05-12
d3q27_psm development

### LINKED ISSUE #282: Bug in d3q27_PSM TRT
If I'm not mistaken in:
https://github.com/CFD-GO/TCLB/blob/3e4154bf071009071f79c06bbe67a6d4a7cddda5/models/particles/d3q27_PSM/Dynamics.c.Rt#L945
`uPsq` is used but not set (to sum of squares of uP)

@ndivaira this can be the reason for weird TRT behavior you noticed.

llaniewski:
Further development of the d3q27 partially saturated cell method model (PSM).

## Problems to adress:
- [x] Very slow code generation
- [x] Legacy variables
  - [x] `fdel`
  - [x] `localGravitationY`
  - [x] `Gravitation#` and `Accel#` duplicate functionality
  - [x] `localOmegaF`
- [x] Lack of units for some settings and globals
- [x] integration of non-conserved variables (`Velocity` not `Momentum`)
- [x] TRT not properly working (maybe #282)
- [x] moment-based implementation could be faster then BGK-style
- [x] *slow*

## Currently
For now, this is just a rewrite of the equilibrium calculations and getting rid of some of the R loops to speed up the code generation.
### Minor
- fixes #282
- deletes `fdel`
- moves the redudant `NonEquilibriumExtrapolation...` code to a function
- changes `TotalFluidVelocity...` to `TotalFluidMomentum...` and adds `TotalFluidMass` globals


llaniewski:
@ndivaira  FYI My developments will appear here. I'll have a working moment-based implementation probably tomorrow for you to test.

ndivaira:
Fantastic. It seems quite likely that not setting uPsq is the cause of the TRT error, but I managed to skip over it.

I'm planning on coming into the office tomorrow, I might see you then.

Cheers,

Nathan
________________________________

ndivaira:
Correction of 'feqP' in CollisionTRT_PSM has fixed bug in TRT. Issue #282 can be closed. For single settling sphere test, fixed and moving reference frame now closely match. However, error still increases as Re increases:

TRT_error

llaniewski:
@ndivaira I had a big delay, because I was getting wrong convergence results. Finally I identified the problem and patched it in #287 

The current stage include moment-based implementation for both collision and `CalcF`. It also drops back to standard BGK/TRT if you don't select any PSM method (NEBB/SUP). It implements TRT in the moment-space also.

llaniewski:
The speed was improved by around 1/3. But it's still far from `d3q27_cumulant`, which is very weird.

ndivaira:
Looks great, as you've been updating I've been running some test cases on Magnus (cpu) with the updated model, which all seem to be working. The speed increase is quite noticable.

PR #292, opened 2020-06-19
Generate model .R file for debuging - no important performance penalty

mdzik:
Generate .R file for Dynamics.c.Rt - it is sourcable for Rstudio, and it make model development much easier :)


llaniewski:
It's a good idea, but I think that it would be better as an `./configure` option.

I added `-k`/`--keep-code` option to `RT` (https://github.com/llaniewski/rtemplate/pull/1) and we can activate this option in a similar way to https://github.com/CFD-GO/TCLB/blob/master/src/configure.ac#L663

I also corrected the `code-fallback` to also kick in (save source), when RT failed to run.

llaniewski:
Solved by #299

Issue #213, opened 2019-09-11
Using SHA checksum for testing results is insustainable


llaniewski:
As we go into testing on multiple platforms (cuda/cpu, different compilers etc), the comparison based on comparing checksums of `vti` files cannot work. The result files have to be compared as numerical fields, up to a certain precision.

Currently running the `tools/tests.sh` fails on most architectures, as the sha sums are from Travis-CI VMs.

Possible solutions:
- base comparison on absolute difference between fields below some epsilon, done with:
   - `vtk` modules in `python` (lot of potential compatibility issues there)
   - some quick write up of a simple `vti` loader for `R`/`python` (not based of vtk, and no support for other formats)
   - `pvpython`?
- base comparison only on other types of output (TXT, Log, etc)

ggruszczynski:
We have faced the same problem recently.
I think that the `tools/tests.sh` shall call a python script which would compare output and expected .vti files, ie. Solution 1a). Compatibility issues occur with g++, nvcc, etc as well ;-)
They can be mitigated installing a **predefined** environment, like
```
$ apt-get install python3.6
$ pip3 install somepackage==1.2.3
```
Having the data read into an array, we can _assert_allclose_ similar to:
https://github.com/CFD-GO/TCLB_tools/blob/master/Python/symbolic_tools/tests/test_vti_reader.py

mdzik:
I would say that test's that run on build machine are too short (to few iterations) to use "epsilon" - we could miss something. I could rewrite the test to include the platform(whole toolchain in fact) in test reference. So CUDA will be tested against CUDA.sha etc. And the author needs to provide the reference "by hand"/"by script".

My idea was that you test manually if solver work (convergence, reference solution etc.) and then you generate ci-test that covers ~max 1 minutes of computations and store the result. If anything changes in output manual reevaluation is needed. The problem is that repo structure does not handle it well and we lack scripts to handle that.

And there is still option to not include VTI just leave CSV as tests.

My proposition would be to create ./test.sh [model] [platform] [mode] script were:

- platform if predefined based on Travis input?
- mode is short/long where it changes number of iterations used in .xml (so, in fact, xml.Rt or something similarly crazy ;) )

I think that I would also need to provide script to run test env (travis) locally which is perfectly doable

And I would vote to include "tests" in TCLB main repo - most repos have it like that.

mdzik:
[...]

PR #303, opened 2020-08-03
Moving all tests to "by value" approach

llaniewski:
# Problem
In #213 I postulated that comparing results based on SHA1 sums is not a good long term strategy.

The current system makes:
- testing on different architectures impossible
- contributing new test hard

# Solution
#259 introduced comparison of vti files based **on value** not checksum. This pull request removes the checksum tests entirely.

## Details
I removed the checksums for vti files, and kept just the last vti file for every test. 

## Heavy tests
This makes `TCLB_tests` repository potentially heavy in the future. We'll try to deal with it as it comes.

### Potential for GIT LFS
The `git lfs` adds a useful interface for tracking heavy files in git. The problem is that github imposes 1GB storage and 1GB/month bandwidth cap. The bandwidth especially would run out quickly combined with our continuous integration (tests on Travis-CI).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "TCLB_tests"]
 	path = tests
-	url =  https://github.com/CFD-GO/TCLB_tests.git
+	url =  ../../CFD-GO/TCLB_tests.git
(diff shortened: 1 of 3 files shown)


llaniewski:
@mdzik Your two phase models are very fragile to change of architecture.

`d2q9_kuper` gives a difference of around 16*[machine epsilon] on `Rho`. This is acceptable and we can change the tolerance on this test.

But both in `d2q9_pf` and `d2q9_pf_fd` there is a `DEBUG`/`Normal` field which gives very different results on GPU and CPU. 6 or even 10 orders of magnitude above machine precision.

mdzik:
- I'll check pf models - this is a bit strange, please disable those tests. I will refactor and clean those models next
- Kuper model is "obsolete". Please change tolerance for now, or disable it. I will try to make PR today

llaniewski:
@mdzik Thanks for the update.

PR #307, opened 2020-08-17
Generic SCMP model 

mdzik:
I added the model with different collision models and different SCMP approaches, for now: 2.
The intended purpose is to accumulate different models in the same framework to ease comparison and evaluation in a particular use case.

The model was tested and appears to be working as intended (alpha;)). Not all options combinations are useful and tested and might lead to unworking code - some pretest will be added later.

The d2q9_scmp_kuperstokh_* models shall recover d2q9_kuper fully


llaniewski:
@TravisMitchell If you will have a time, I would appreciate you looking into this development. I made you a reviewer.

llaniewski (approved):

PR #324, opened 2020-09-24
+ d3q27_pf_velocity contact angle fixes

TravisMitchell:
- I created explicit variable rather than relying on implicit double to int conversion in Solid Normal calculation.
- I also removed the reliance on integer division in defining values for the solidFlag variable in normal calculation.
- I added a "fix" for edge condition where a solid normal points towards another solid cell. A more robust fix is needed, but this stops the simulation breaking. Was an issue I was finding when importing STL of randomly backed, arbitrary shaped objects from a DEM code I was using.

### DIFF
--- a/models/multiphase/d3q27_pf_velocity/Dynamics.R
+++ b/models/multiphase/d3q27_pf_velocity/Dynamics.R
@@ -161,7 +161,7 @@ if (Options$thermo){
 }
 
 # Stages - processes to run for initialisation and each iteration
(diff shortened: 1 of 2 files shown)


TravisMitchell:
Please hold off on merging this please Lukasz, doesn't seem to work on Gadi.... Will try to fix at the start of next week.

llaniewski:
@TravisMitchell you can mark a pull request as "draft". And when you finish just ask for a review.

PR #340, opened 2020-11-18
Adding PercentChange to Stop

llaniewski:
This PR adds and option to stop computation when **percent** change of some global is below some level.

Usage:
```xml
<Stop FluxPercentChange="1%" Times="10" Iterations="10"/>
```
**reminder:** in TCLB `%` is a "unit" equivalent to `0.01`. So you can say `"1e-3"` or `"0.1%"`.

FYI: @TravisMitchell


TravisMitchell:
Almost the same as what I had put into my local copy.. Didn't realise the % was a unit, so had just worked it off user inputting x%. That is good to know, thank you!

llaniewski:
@TravisMitchell Added your suggested changes.

As for useful units, we have "%" - percent, "d" - degrees (converts "90d" to pi/2). and Avogadro number which I just noticed we have to fix, as it's named "An" for some reason :D

PR #343, opened 2021-01-20
Refactoring of the LatticeBase, and fixes to arbitrary grid.

trisbsamson:
Fixed bugs in computing maximum block sizes for arbitrary lattice kernels.

Removed some redundant code leftover from moving some functions into parent classes/files.

Moved all functions which have identical or similar functionality in ArbitraryLattice.cu and Lattice classes to LatticeBase.cu; these were generally functions used to save/load FTabs, those which handle recording snaps and keeping track of snap numbers, and some settings functions which don’t depend on model-specific information generated by the R-template.

Removed redundant variables left over in ArbitraryLattice.cu and ArbitraryLatticeContainer.cu and fixed errors which this created.

Moved the calculation of CUDA grid size from the ArbitraryLatticeContainer init function into the function which calls the kernel. This meant I could directly use the ThreadNumber::getSY() value computed for the correct node type and kernel call.

Created another macro in cross.h to access the CUDA variable gridDim, which i’m using to compute the node ID inside the kernel.

Created another couple of macros in cross.h to support CPU computation; similar idea to what already exists for the Cartesian lattice, just suiting my grid layout.


llaniewski (approved):
I corrected the averages initialization in d3q27_cumulant, and now it passes the tests.

PR #359, opened 2021-07-09
Feature checkpoint

TravisMitchell:
I added checkpoint feature to improve the usability of this function. User specifies how many checkpoints to keep saved, the default is one. A restart xml file is created that has a loadBinary element added prior to the first solve element.

### DIFF
--- a/src/Handlers/cbSaveCheckpoint.cpp
+++ b/src/Handlers/cbSaveCheckpoint.cpp
@@ -20,6 +20,11 @@ int cbSaveCheckpoint::Init () {
 			} else {
 				// If the attr value is not the string all, assume it is an int
 				keep = attr.as_int();
+				if ( keep < 0) {
+					// Check the user hasn't set keep to a negative value
+					error("Keeping a negative no. of chckpnts not allowed, returning default behaviour.");
+					keep = 1;
+				}
 			}
 		} else{
 			keep = 1;
--- a/src/Handlers/cbSaveCheckpoint.cpp
+++ b/src/Handlers/cbSaveCheckpoint.cpp
@@ -62,12 +62,14 @@ int cbSaveCheckpoint::DoIt () {
 			if (myqueue.size() > keep) {
 				// myqueue should only ever reach the size of keep
 				fileStr = myqueue.front();
-				remove( fileStr.c_str() ); //Takes char
+				int rm_result = remove( fileStr.c_str() ); //Takes char
+				if (rm_result != 0) error("Checkpoint file was not deleted: %s",fileStr);
 				myqueue.pop();
 
 				if (D_MPI_RANK == 0 ) {
 					restStr = myqueue_rst.front();
-					remove( restStr.c_str() );
+					rm_result = remove( restStr.c_str() );
+					if (rm_result != 0) error("Restart file was not deleted: %s",restStr);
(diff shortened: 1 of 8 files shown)


src/Handlers/cbSaveCheckpoint.cpp
```diff
@@ -0,0 +1,98 @@
...
+				// Look for the keyword all and use keep = 0 as flag to store all
+				keep = 0;
+			} else {
+				// If the attr value is not the string all, assume it is an int
+				keep = attr.as_int();
+			}
+		} else{
+			keep = 1;
+		}
+
```
llaniewski, line 27: Please add a sanity check `keep >= 0`

src/Handlers/cbSaveCheckpoint.cpp
```diff
@@ -0,0 +1,98 @@
...
+		char filename[2*STRING_LEN];
+		std::string fileStr;
+		std::string restStr;
+
+		solver->outIterCollectiveFile("checkpoint", "", filename);
+		solver->outIterCollectiveFile("restart", ".xml", restartFile);
+		
+		fileStr = solver->lattice->saveSolution(filename);
+
+		myqueue.push( fileStr );
```
llaniewski, line 51: push to the queue only when `keep != 0`, otherwise we'll be filling up an unnecessary memory.

src/Handlers/cbSaveCheckpoint.cpp
```diff
@@ -0,0 +1,98 @@
...
+			writeRestartFile(filename, restartFile);
+			restStr = restartFile;
+			myqueue_rst.push( restStr );
+		}
+
+		if (keep != 0){
+			if (myqueue.size() > keep) {
+				// myqueue should only ever reach the size of keep
+				fileStr = myqueue.front();
+				remove( fileStr.c_str() ); //Takes char
```
llaniewski, line 62: It's always good to handle the errors. So: catch the output value of remove, and print error if something goes wrong.

src/Handlers/cbSaveCheckpoint.cpp
```diff
@@ -0,0 +1,98 @@
...
+				myqueue.pop();
+
+				if (D_MPI_RANK == 0 ) {
+					restStr = myqueue_rst.front();
+					remove( restStr.c_str() );
```
llaniewski, line 67: same as above

src/Handlers/cbSaveCheckpoint.cpp
```diff
@@ -0,0 +1,98 @@
...
+			}
+		}
+
+		return 0;
+	};
+
+int cbSaveCheckpoint::writeRestartFile( const char * fn, const char * rf ) {
+
+		// Check if a LoadBinary attribute exists
+		pugi::xml_node n1 = solver->configfile.child("CLBConfig").child("LoadBinary");
```
llaniewski, line 79: I think this should be done on a copy of the XML tree structure. Otherwise you can run into big trouble, as the code is still traversing this XML tree.

For example: if this is called early in the TCLB run, and it adds this "LoadBinary", TCLB will go further, and encouter this "LoadBinary" further along the way.

llaniewski (changes requested):

llaniewski (approved):

PR #362, opened 2021-08-01
On recovering 2nd order convergence of lbm with reaction type source terms

ggruszczynski:
I added the model & example described in <https://arxiv.org/abs/2107.03962>


llaniewski:
Please change this pull request to merge with `develop` not `master`

PR #376, opened 2021-10-29
HIP support

llaniewski:
This PR introduces preliminary support for compilation with HIP.

To compile for the ROCM architecture, use
```bash
make configure
./configure --enable-hip
```

The whole code is compiled in this case with `hipcc` (nothing with `mpic++`), which means all library paths have to be discovered.

FYI: @mdzik @Me-Basha 

## Notes
*This is preliminary.*
### TODO
- Tuning and testing the different warp size. Everything now is made for 32
- Moving to a more reasonable naming convention, as currently everything has "Cuda" in it.


llaniewski:
@mdzik I introduced some changes to the hip version which reduce the number of memcpy-s done while importing geometry and exporting quantities. This was a major bottleneck on MI150 (don't know about MI250)

mdzik:
I will benchmark it on LUMI today and let you know. There are also
bottlenecks in the previous version of the branch on MI250s, so we will see.
It is also advised to use GPU pointers in MPI communication - do you think
its doable?

Pilot phase was extended till 8 or 10th of january, we could schedule some
profiling time if you have some?

pon., 5 gru 2022, 07:55 użytkownik Łukasz Łaniewski-Wołłk <
***@***.***> napisał:

PR #378, opened 2021-11-17
Fix d3q27 pf corner bug

TravisMitchell:
A few changes and updates to be pushed:
- minor updates in flow d3q27 and d3q27_cumulants to make perm calculations easier
- minor updates in d2q9_pf_velocity to allow for droplet on solid in shear flow test case
- clean up of d3q27_pf_velocity and attempted fix of corner bug created with the q15 lattice for the allen-cahn equation
- also added required outputs to get densities for machine learning applications.
- grabbed from Bryce's branch the addition of BOUNDARY field for catalyst operations
- added d3q27_n1 values to globals from Dmytro's branch for alternate contact angle implementation


TravisMitchell:
will resolve these conflicts, a bit later before talking to @llaniewski about merging.

llaniewski:
@TravisMitchell how this is going?

TravisMitchell:
@llaniewski I just resolved the conflicts and it should be okay to merge for now. It has some improvements on the current d3q27_pf_velocity model that is in develop, and the test gave back equal results in the benchmark repo that I have for the model.

PR #382, opened 2022-01-10
Fix make for all options available in d3q27_pf_velocity

TravisMitchell:
last commit broke some of the lesser used options in the model: d3q27_pf_velocity.
this commit fixes this.


llaniewski:
@TravisMitchell I don't think it's a good idea to push the ML related things to the main repo.

PR #390, opened 2022-04-27
D2q9 reaction diffusion configurable upgrade

mdzik:
I modified the model to support different time-integration models. 
Keep in mind that trapezoid is stable, others should be considered experimental (they converge and work, but no guaranties)


llaniewski:
@mdzik, I'm all for supporting material, but this pull request includes an executable `a.out` :/

mdzik:
Sorry, one "git add" to far.

PR #397, opened 2022-05-22
Export iteration number to Rinside

shkodm:
Makes much easier generating files associated with the current iteration (e.g. I used it to generate pictures of "slices" data and saving into files with filename based on the iteration)

Not sure, where exactly it should be, I put it into `Solver$Globals` , because it made more sense than `Solver$Iteration$`, but suggestions are welcome


llaniewski (changes requested):
I think it being in "Globals" is good. Please add it also to the `Names()` function, so it will be visible in autocomplete, etc.

shkodm:
@llaniewski Done

llaniewski (approved):

PR #398, opened 2022-05-25
Adding force averaging to simplepart

llaniewski:
Adding options to force averaging to simplepart (off by default). To use:
```xml
<Log name="out.csv" Iterations="64" average="y"/>
```

FYI @TravisMitchell


TravisMitchell:
thanks @llaniewski, I think this will work for the particle bed case quite nicely

PR #400, opened 2022-06-22
+ add model using lbmpy for collision rules and examples

TravisMitchell:
I added a d2q9 cumulants model using lbmpy to generate initialisation and collision kernel.
I added examples that correspond to other 2D single phase tests.

This provides an example for anyone that is looking to make a new model that may already exist in lbmpy, but wants the parallelisation available in tclb. Note that python libraries lbmpy and pystencils are required for this model.


ggruszczynski:
excellent!

llaniewski:
It looks very interesting. It would be maybe good to be able to do some preparation in Dynamics.R, to guarantee that the stencil matches the streaming directions.

Sidenote: does `lbmpy` do any boundary conditions?

TravisMitchell:
it does have boundary conditions, but I need to look into where they set them up. I just thought this would be quite interesting for a few people, so wanted to push through a minimal working example for generation of initialisation and collision :).

I completely agree with the preparations in Dynamics.R, and I will look into doing this as I progress.

There are also additional optimisations that I can make use of for minimising floating point operations and these types of things.

PR #402, opened 2022-07-05
Fix leak in Sampler

shkodm:
During using Sampler I realised that added points are not cleared on finalising. This means next time `Sampler` is used (e.g. if used inside  two successive `Solve` handlers) previous points from the previous Sampler would still be there and would be unexpectedly present in the output.

This fixes it.


shkodm:
@llaniewski now that you are back, you can merge / review this

llaniewski (approved):

Issue #416, opened 2022-11-30
d3q27_pf_velocity has duplicate function in develop branch.


TravisMitchell:
Opening issue to match associated pull request to remove function and add additional boundary capability

PR #417, opened 2022-11-30
Fix current issue #416 and addition of extra boundary conditions for d3q27_pf_velocity

TravisMitchell:
Fix for #416 
Addition of pressure/velocity boundary conditions for all directions.


models/multiphase/d2q9_pf_velocity/Dynamics.R
```diff
@@ -141,8 +144,10 @@ AddGlobal(name="InletFlux", comment='pressure loss', unit="1m2/s")
 AddGlobal(name="TotalDensity", comment='Mass conservation check', unit="1kg/m3")
 AddNodeType(name="SpikeTrack", group="ADDITIONALS")
 AddNodeType(name="BubbleTrack", group="ADDITIONALS")
+AddNodeType(name="WAVE", group="ADDITIONALS")
```
shkodm, line 147: Since other node types are not all caps, I would expect this one to be just "Wave" for consistentcy

models/multiphase/d2q9_pf_velocity/Dynamics.R
```diff
@@ -99,6 +100,8 @@ AddSetting(name="Period", default="0", comment='Number of cells per cos wave')
 AddSetting(name="Perturbation", default="0", comment='Size of wave perturbation, Perturbation Period')
 AddSetting(name="MidPoint", default="0", comment='height of RTI centerline')
 
+AddSetting(name="wave", default="0", comment='RTI (=0) vs wave (>0) switcher')
```
shkodm, line 103: Why all lowercase? Here I would also prefer just "Wave"

models/multiphase/d3q27_pf_velocity/Dynamics.R
```diff
@@ -213,7 +214,17 @@ if (Options$altContactAngle){
 	AddNodeType(name="MRT", group="COLLISION")
 	if (Options$OutFlow){
 		AddNodeType(name="ENeumann", group="BOUNDARY")
+		AddNodeType(name="WNeumann", group="BOUNDARY")
 		AddNodeType(name="EConvect", group="BOUNDARY")
+		AddNodeType(name="WConvect", group="BOUNDARY")
+		## AddNodeType(name="NNeumann", group="BOUNDARY")
```
shkodm, line 220: If you need those, can't you add them in the same way (with a loop), as you did for pressure and velocity boundaries?

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
```diff
@@ -238,27 +223,20 @@ CudaDeviceFunction void calcPhaseGrad_init(){
...
-	cat(paste0("\t return g",myDensity,";\n"))
-	cat("}\n")
-}
-for (myDensity in 0:14){
-	cat(paste0("CudaDeviceFunction real_t geth",myDensity,"(){\n"))
-	cat(paste0("\t return h",myDensity,";\n"))
-	cat("}\n")
-}
+###################################################
+# get functions to define outputs indicated in .R #
```
shkodm, line 231: `get` -> `Get`

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
```diff
@@ -274,21 +252,26 @@ CudaDeviceFunction vector_t getU(){
...
 }
 CudaDeviceFunction vector_t getNormal(){
 	vector_t n = {nw_x, nw_y, nw_z};
 	return n;
 }
 
-//######HELPER FUNCTIONS#######//
+<?R
+###################################################
+# miscellaneous helper functions used in the code #
```
shkodm, line 269: `miscellaneous` -> `Miscellaneous`

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
```diff
@@ -320,11 +314,17 @@ CudaDeviceFunction vector_t calcGradPhi(){
 }
 
 CudaDeviceFunction void calcPhaseF(){
+	// This function is run as the second stage of each action and fills
```
shkodm, line 317: I think docstrings for C/C++ functions are conventionally placed above the function they are describing with /* */ as comments. So this should be (also applies to other function where docsting was added) 

```cpp
/* This function is run as the second stage of each action and fills the iteration
buffer with updated values of PhaseF */
CudaDeviceFunction void calcPhaseF(){
     ....
```

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
```diff
@@ -938,77 +1009,164 @@ CudaDeviceFunction void ENeumann(){
...
-		n = c(-1,0,0)
-		heq = pf * EQ_h$feq
-		bounce = Bounce(U_PF)
-		sel = as.vector( (U_PF %*% n) < 0)
+}
 
-		sel2 = as.vector( ( U_PF %*% n) == 0)	
-		exM = (h[sel2] - heq[sel2]) %*% rep(1,sum(sel2)) 
-		Unknowns = 1.0/sum(sel)
+/* If needed, user can uncomment and adjust code for these B.C.
```
shkodm, line 1021: I think you should be able to generate those with R as well, similarly as you do for the pressure/velocity boundaries. Otherwise there is quite a lot of duplicated and commented out code, which is not nice

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
```diff
@@ -1323,24 +1414,18 @@ CudaDeviceFunction void updateBoundary(){
...
+			<?R C(my_velocity_boundaries[ii])?>();
 			break;
-		case NODE_WPressure:
-			WPressure();
+		case NODE_<?R C(my_pressure_boundaries[ii])?>:
+			<?R C(my_pressure_boundaries[ii])?>();
 			break;
+		<?R
+        }
+        ?>
```
shkodm, line 1428: Convective boundaries should be also handled here, right? Or at least a comment should be add, that this should be adjusted if certain convective boundaries are uncommented

models/multiphase/d3q27_pf_velocity/Dynamics.R
```diff
@@ -6,17 +6,6 @@ if (Options$q27){
 	source("d3q27q15.R")
 }
 
-if (Options$ML){
```
shkodm, line 9: Was this intended  to be removed? Why? The same applies for 3D model

shkodm (changes requested):
@TravisMitchell I reviewed

models/multiphase/d3q27_pf_velocity/Dynamics.R
```diff
@@ -62,8 +51,18 @@ if (Options$altContactAngle){
 if (Options$OutFlow){
 	for (d in rows(DensityAll)) {
 		AddField( name=d$name, dx=-d$dx-1, dy=-d$dy, dz=-d$dz )
+		AddField( name=d$name, dx=-d$dx+1, dy=-d$dy, dz=-d$dz )
+		## AddField( name=d$name, dx=-d$dx, dy=-d$dy-1, dz=-d$dz )
```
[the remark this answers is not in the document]
TravisMitchell: I have removed these to avoid confusion. A comment has been added to .c.Rt that if for some reason more outflow conditions that aren't E or W are required - then they need to be implemented by the user using existing code as template

models/multiphase/d3q27_pf_velocity/Dynamics.R
TravisMitchell: the extra fields required slow compute and also need to be included in gradient calculations. I have removed these as per above.

models/multiphase/d3q27_pf_velocity/Dynamics.R
TravisMitchell: decided that the ML was rather a specific niche thing currently used for testing and not ready for develop branch

shkodm (approved):
Looks good to me.

Issue #186, opened 2018-09-03
VTK for subregion


llaniewski:
Add capability to specify a region in `<VTK/>` export.

PR #426, opened 2023-03-20
VTK and HDF5 output additions + cpu output fix

bhill23:
Addresses #186. Region definition works similarly to Geometry definition, negative dx and nx work backwards from domain end.
Also adds nodetypes to HDF5 outputs.
Fixes outputs for cpu only builds where the z axis was not being looped over.


llaniewski (approved):

llaniewski:
Fixes #186

PR #431, opened 2023-04-06
Add wetting boundary conditions to the phase field model

shkodm:
- I modified the model `d3q27_pf_velocity`  (maintainer: @TravisMitchell ) to include various variants of the wetting boundary conditions.
- I also added helpful export of `OutputPath` (and maybe other info) to Rinside.

I also splitted `Dynamics.c.Rt` in two files with all wetting boundary conditions code in `Boundary.c.Rt` to make it more readable. The file is included and preprocessed using Rtemplate <?RT ?> syntax at the bottom of `Dynamics.c.Rt`. 

The code in `Boundary.c.Rt` should be ready to review already, @TravisMitchell, would be useful if you go through. It should be quite documented. 
I need to go through the code in `Dynamics.c.Rt` and `Dynamics.R` again to make sure I did not remove / added something by accident when cleaning up code and fixing merge conflicts. I also want to double check the performance again just in case. This is why for now PR is DRAFT, once I verify this I will convert it into a normal PR.


src/conf.R
```diff
@@ -569,7 +569,10 @@ Fields$tangent_name = add.to.var.name(Fields$name,"d")
 
 Fields$area = (Fields$maxx-Fields$minx+1)*(Fields$maxy-Fields$miny+1)*(Fields$maxz-Fields$minz+1)
 Fields$simple_access = (Fields$area == 1)
-Fields$big = Fields$area > 27
+# NOTE: This heuristic does not always work, currently enabled only for fields
+# which have more dynamic access pattern
+# Fields$big = Fields$area > 27
+Fields$big = ifelse(Fields$name %in% c("gradPhiVal_x", "gradPhiVal_y", "gradPhiVal_z", "gradPhi_PhaseF"), TRUE, FALSE)
```
llaniewski, line 575: This has to be non-model specific. Best if it could be selected some way at "AddField/AddDensity".

models/multiphase/d3q27_pf_velocity/model.R
```diff
@@ -0,0 +1,86 @@
+<?R
```
llaniewski, line 1: This is a Rt file, but the suffix is ".R". If this is pure R, just get rid of `<?R` stuff.

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
```diff
@@ -1669,6 +1439,11 @@ CudaDeviceFunction void updateTrackers(real_t C){
 		return lower + current*(upper - lower);
 	}
 #endif
+
+
+/* Include file with the wetting boundary condition code  */
+<?RT models/multiphase/d3q27_pf_velocity/Boundary.c.Rt ?>
```
llaniewski, line 1445: 1. it would be good for if  `rtemplate` would search for the included RT files the same way it searches for `source(...)` files. Here is the source mechanics: https://github.com/llaniewski/rtemplate/blob/master/R/RTemplate.R#L13
2. the main makefile has to discover some way that this file pulls the other file. Otherwise it will not regenerate `Dynamics.c` when `Boundary.c.Rt` is modified.

llaniewski (changes requested):

models/multiphase/d3q27_pf_velocity/Dynamics.R
```diff
@@ -6,53 +6,76 @@ if (Options$q27){
...
-    AddField("PhaseF",stencil3d=3, group="PF")
 
-    AddStage("WallInit_CA"  , "Init_wallNorm", save=Fields$group %in% c("nw", "debug_boundary"))
-    AddStage("calcWall_CA"  , "calcWallPhase", save=Fields$name %in% c("PhaseF", "der_tangent_1_wall", "der_tangent_2_wall", "TangentWallVector1_x", "TangentWallVector2_x", "TangentWallVector1_y", "TangentWallVector2_y", "TangentWallVector1_z", "TangentWallVector2_z", "perpVal"), load=DensityAll$group %in% c("nw", "gradPhi", "PF"))
+AddDensity("IsSpecialBoundaryPoint", dx=0, dy=0, dz=0, group="solid_boundary")
+AddQuantity("SpecialBoundaryPoint", unit = 1)
 
-    AddStage('calcPhaseGrad', "calcPhaseGrad", load=DensityAll$group %in% c("g","h","Vel","nw", "PF"), save=Fields$group=="gradPhi")
-    AddStage('calcPhaseGrad_init', "calcPhaseGrad_init", load=DensityAll$group %in% c("g","h","Vel","nw", "PF"), save=Fields$group=="gradPhi")
+if (Options$altContactAngle){
```
TravisMitchell, line 61: Could we change the option name from _altContactAngle to _geometric?

models/multiphase/d3q27_pf_velocity/model.R
shkodm: Nice catch, will do

src/conf.R
shkodm: Yes, I know. I was planning to add additional flag to AddDensity / AddField

models/multiphase/d3q27_pf_velocity/Dynamics.R
shkodm: Oh, yes, I completely forgot. The names became too familiar in my mind.

src/conf.R
shkodm: @llaniewski, do we want to keep the default setting `Fields$big = Fields$area > 27` ?  

I also am going to change the name of this settting from `big` to `optimise_for_static_access`, unless you have a better name suggestion

llaniewski:
I think the `optimise_for_static_access` is not working right for the 2D models. the "big" flag was triggered for stencils above 27, which never happened for 2D models, and thus, this code was never generated for 2D models. Now that you generate it, it fails on compilation. I think this comes from an error (wrong assumption) in `conf.R`. Try replacing (https://github.com/CFD-GO/TCLB/blob/develop/src/conf.R#L777):
```R
tab3 = rep(TRUE,9)
```
with one which will take into account what are the maximal stencils in each direction:
```R
tab3 = c(mins<0,TRUE,TRUE,TRUE,maxs>0)
```

models/multiphase/d3q27_pf_velocity/Dynamics.c.Rt
shkodm: 1. Yes, I opened PR for Rtemplate
2. I adjusted the makefile, but on the way I also realised, there are more issues: 
   -  Since `Dynamics.R` can source some R files (as Travis is already doing in his own model without this PR even and  maybe others are doing as well), any change to those files could potentially alter the model and therefore it has to be rebuild. Currently only changes to `Dynamics.R` trigger such rebuilds. I changed that to be all `.R` files in the model directory instead. This is not ideal, as at will have some false positives, but I think this is better than building some kind of dependency graph of R files based on how they are sourcing each other 
   - Similarly Dynamics.c now depends on other .Rt files in the model directory

shkodm:
> I think the optimise_for_static_access is not working right for the 2D models. the "big" flag was triggered for stencils above 27, which never happened for 2D models, and thus, this code was never generated for 2D models. Now that you generate it, it fails on compilation. 

I will fix the issue with 2D models, but actually `big` flag was not supposed to be triggered for 2D model. `Big`  means no optimisation (so corresponds to `optimise_static_access = FALSE`). I had it the other way around, that is why new 2D models build was triggered and was failing.

shkodm:
@TravisMitchell you can review now. I already tested the performance (everything is still god), will test again if the results are the same as I had before.

shkodm:
@TravisMitchell would you have time to review this one soon? I validated it on capillary intrusion already, the droplet spread jobs are currently stuck in queue.

TravisMitchell:
@shkodm I will do today

shkodm:
@TravisMitchell I tested all methods on the Washburn Law again after your refactoring, it still works. I think you can approve this PR. 

@llaniewski I think it can be ready to merge. But please squash my commits when merging (there should be an option), there is no need for my 100 commits to be in the git history of the main repo.

Also I think to keep: 
```
<?RT models/multiphase/d3q27_pf_velocity/Boundary.c.Rt ?> 
```
as it is (not using just `Boundary.c.Rt`), because this way we don't need to update `rtemplate`  package everywhere.

llaniewski (approved):

TravisMitchell (approved):

PR #442, opened 2023-08-18
Fixing the SolidGrid print issue

llaniewski:


llaniewski:
@ndivaira This should solve the issue with excessive printing and reconstruction of the SolidGrid indexer. I'll merge after all the tests complete.

Issue #437, opened 2023-07-26
d3q27_pf_velocity - read and write stencil fields in same stage


TravisMitchell:
There are currently a couple issues with the d3q27_pf_velocity model that need to be resolved. These do not necessarily impact results, but are unsafe access behaviours that should be amended:
1. Currently the wall phase update reads PhaseF and writes PhaseF. This creates potential race condition as it is a stencil field. To amend this, the solid flag field could be modified to wall phase field value and updated from PhaseF in the wall phase update.
2. Init_distributions currently saves and writes PhaseF. This doesn't look to be necessary, but could create issues in e.g. boundary cells.

@shkodm this might concern you. But I will either fix when I have time or if you have time feel free to also fix.

PR #444, opened 2023-08-28
Field access checks

llaniewski:
This PR introduces a new mechanism for setting and testing the access policy for fields in stages.

## Setting access
The `AddStage` has now three arguments:
- `load.densities=` specifies the `Densities` that are loaded at the beginning of the stage.
- `read.fields=` specifies the `Fields` that can be accessed in the Dynamics with `field(dx,dy,dz)` macros.
- `save.fields=` specifies the `Fields` that are saved at the end of the stage.

### Defaults
By default the actions/stages setup is:
```R
AddStage(main="Run", name="BaseIteration", load.densities=TRUE, read.fields=TRUE, save.fields=TRUE)
AddStage(main="Init", name="BaseInit", load.densities=FALSE, read.fields=FALSE, save.fields=TRUE)
AddAction(name="Iteration", stages=c("BaseIteration"))
AddAction(name="Init", stages=c("BaseInit"))
```
Moreover, the default `read.fields` is as allowing as possible. If not specified otherwise it is assumed that stage has read access to all fields that are available to it.

## Checks
The access policies are checked in each `Action` (series of `Stages`) assuming:
- You cannot read a field which wasn't written.
- You cannot overwrite already written fields
- You should write all fields by the end of the action
- `Init` action should not read anything from previous iteration.

### Exceptions
There are three mechanisms to circumvent checks:
- `non.mandatory=TRUE` in `AddField` signals that this field doesn't have to be written by the end of an action
- `can.overwrite=TRUE` in `AddStage` signals that a stage can overwrite fields
- `SetOptions(permissive.access=TRUE)` switches the test entirely and sets all fields to be readable in all stages. **This should be avoided**

## Modifications
Some adjustments were made to models to agree with these policies. All models were tested and `permissive.access=TRUE` was added to the ones failing from this change. These models should be fixed to accommodate these checks. Especially `d3q27_pf_velocity`, as per #437.

### DIFF
--- a/models/electrokinetic/d2q9_poison_boltzmann/Dynamics.R
+++ b/models/electrokinetic/d2q9_poison_boltzmann/Dynamics.R
@@ -1,3 +1,10 @@
+# Setting permissive access policy.
+#  * This skips checks of fields being overwritten or read prematurely.
+#  * Otherwise the model compilation was failing.
+#  * This should be removed if the issue is fixed
+SetOptions(permissive.access=TRUE)  ### WARNING
+
+
 
 AddDensity( name="g[0]", dx= 0, dy= 0, group="g")
 AddDensity( name="g[1]", dx= 1, dy= 0, group="g")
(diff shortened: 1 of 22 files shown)


llaniewski:
I added feature for plotting the stages and their access to a pdf file. The plot shows the field read and write status and marks problems in red. When you run
```bash
make d3q27_pf_velocity/summary
```
this nice plot is generated:
Screenshot_2023-08-30_16-22-16

FYI: @TravisMitchell @shkodm @mdzik

PR #455, opened 2023-10-19
Constant memory refactor

kubagalecki:
GPU data pointers stored in `LatticeData` and similar classes have thus far been stored in constant device memory and manually updated when needed. The idea is to now pass them directly to kernels via `const` arguments, which should automatically get placed in constant memory. This significantly simplifies the code and allows for greater flexibility going forward.


kubagalecki:
@llaniewski I think this is nearly everything we discussed. One outstanding item is renaming `GetThreads.h`. My suggestion would be `LaunchUtils.h`.

kubagalecki:
By the way, I ran the benchmark Travis gave me on Gadi, there was a 0.737% slowdown, which I'd write off as noise. This is without `__grid_constant__`. Full log: [taylor-bubble-benchmark.log](https://github.com/CFD-GO/TCLB/files/13060368/taylor-bubble-benchmark.log)

llaniewski (approved):

PR #458, opened 2023-10-26
Adding clang-format configuration and script

llaniewski:
This adds `tools/format.sh` script for formatting C++ files that uses a simple `.clang-format` configuration, and then compacts small `for` loops and multiplications/divisions with `sed`.

FYI @kubagalecki


kubagalecki:
@llaniewski Some notes (I will add more as I think of them):
- `PointerAlignment: PAS_Left`

PR #462, opened 2023-11-20
Arbitrary grid

kubagalecki:
- `toArb` - export Cartesian case as arbitrary
- read .cxn file and init Arbitrary lattice
- partition read lattice using ParMETIS


kubagalecki:
@llaniewski As seen in https://github.com/CFD-GO/TCLB/actions/runs/6926854681/job/18839682950, there was a bug in determining unique directions. I've now fixed it, but it exposed a limitation of HIP: the maximum number of ***unique*** offset directions is 256, e.g. d3q343 would not compile. This is a result of a low limit of fold expression expansions in clang, see [minimum example](https://godbolt.org/z/MGaoeqPoh). There is a way to work around this by writing more R template code, but the question is do we realistically need to support models with q > 256?

kubagalecki:
> There is a way to work around this by writing more R template code

There is also a compiler flag ` -fbracket-depth=N` to raise the limit

PR #468, opened 2023-12-13
RunPython

llaniewski:
This PR extends the RunR capabilities with Python execution through the `reticulate`. You can now execute Python code with `<RunPython/>` element, where `Solver` is exposed in the same way as in R.

### Example
```xml
<?xml version="1.0"?>
<CLBConfig version="2.0" output="output/">
	<Geometry nx="1024" ny="100">
		<MRT><Box/></MRT>
		<WVelocity><Box nx="1"/></WVelocity>
		<EPressure><Box dx="-1"/></EPressure>
	</Geometry>
	<Model>
		<Param name="VelocityX" value="0.05"/>
		<Param name="Viscosity" value="0.01"/>
	</Model>
	<RunR Iterations="100">
		print(sum(Solver$Quantities$Rho))
	</RunR>
	<RunPython Iterations="100">
print(Solver.Quantities.Rho.sum())
	</RunPython>
	<Solve Iterations="1000"/>
</CLBConfig>
```

FYI: @TravisMitchell @shkodm @mdzik 
*I think with this present in the code we can get rid of the native Python support.*

### DIFF
--- a/src/Handlers/cbRunR.cpp
+++ b/src/Handlers/cbRunR.cpp
@@ -737,9 +737,16 @@ int cbRunR::Init() {
 	attr = node.attribute("echo");
 	if (attr) echo = attr.as_bool();
 
+	s_tag++;
+	tag = s_tag;
+	
 	source = "";
-        for (pugi::xml_node par = node.first_child(); par; par = par.next_sibling()) {
+    for (pugi::xml_node par = node.first_child(); par; par = par.next_sibling()) {
 		if (par.type() == pugi::node_element) {
+			if (python) {
+				ERROR("Code-embedded xml nodes not supported for python");
+				return -1;
+			}
 			char nd_name[20];
 			sprintf(nd_name, "xml_%0zx", par.hash_value());
 			R[nd_name] = RunR::wrap_handler(solver,this,par);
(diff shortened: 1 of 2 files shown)


TravisMitchell:
I was already okay with removing native Python support but this also looks nice :).

shkodm:
Very exciting, thanks @llaniewski ! Do you want me to test it using the two-phase fracture setup I was working on?

TravisMitchell:
@shkodm I think I would like you to test on the fracture setup, and maybe add to the multiphase tools repo as a demo for using it ;)

llaniewski:
@shkodm That would be great. Hope this will simplify your setup a bit.

@TravisMitchell @shkodm Some notes:
1. Seems that the read-only part works quite well, but setting stuff back in the `Solver` object doesn't really work well (eg. `Solver.Geometry.BOUNDARY[0,0] = "None"`). I think this would mean much more effort.
2. R objects are auto converted to python. Matrices become numpy arrays, so that's good. But the factor arrays seem to be converted to vectors of strings.

shkodm:
@llaniewski I tested, and it seems to work for my purposes now (plotting phase field on the aperture plot). We can probably merge this one directly now and then create follow up PRs if I find some issues. 

Regarding notes:

1. I don't think I need it right now, but setting stuff back would be cool. Maybe it is possible to do it in the follow up PR? 
2. Factor array converted to strings is actually good, because of https://github.com/CFD-GO/TCLB/issues/470

PR #471, opened 2023-12-20
Arb fix/d2q9 pf velocity

TravisMitchell:
this initial fix at least allows the d2q9 phase field model to be run with normal cases. There is a chance that very complex geometry may cause edge conditions (e.g. where a solid node has a normal direction that points to another solid) that may require the introduction of a new field in the future. 

Note that use of nan makes failcheck require to be applied with a focus area.

### DIFF
--- a/models/multiphase/d2q9_pf_velocity/Dynamics.c.Rt
+++ b/models/multiphase/d2q9_pf_velocity/Dynamics.c.Rt
@@ -221,8 +221,7 @@ CudaDeviceFunction real_t calcMu(real_t myPhase){
 		}
 	#endif
 
-	// eq 5 from "Improved locality of the phase-field lattice Boltzmann 
-	// model for immiscible fluids at high density ratios"
+	// eq 5 from "Improved locality of the phase-field lattice Boltzmann model for immiscible fluids at high density ratios"
 	real_t pfavg = 0.5*(PhaseField_l+PhaseField_h);
 	real_t mu = 4.0*(12.0*sigma/W)*(myPhase-PhaseField_l)*(myPhase-PhaseField_h)*(myPhase-pfavg)
 		        - (1.5 *sigma*W) * lpPhi; 
@@ -230,8 +229,7 @@ CudaDeviceFunction real_t calcMu(real_t myPhase){
 }
 
 CudaDeviceFunction real_t calcGamma(int i, real_t u, real_t v, real_t u2mag){
-	// eq 10 from "Improved locality of the phase-field lattice Boltzmann 
-	// model for immiscible fluids at high density ratios"
+	// eq 10 from "Improved locality of the phase-field lattice Boltzmann model for immiscible fluids at high density ratios"
 	real_t gamma, tmp;
 	tmp = (d2q9_ex[i]*u+d2q9_ey[i]*v);
 
@@ -240,8 +238,7 @@ CudaDeviceFunction real_t calcGamma(int i, real_t u, real_t v, real_t u2mag){
 }
 
 CudaDeviceFunction real_t calcF_phi(int i, real_t numerator, real_t nx, real_t ny){
-	// eq 7 from "Improved locality of the phase-field lattice Boltzmann 
-	// model for immiscible fluids at high density ratios"
+	// eq 7 from "Improved locality of the phase-field lattice Boltzmann model for immiscible fluids at high density ratios"
 
 	// nx: normalized PhaseField gradient in x direction
 	// this method calculates F_phi in [0-9] lattice coord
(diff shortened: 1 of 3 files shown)


models/multiphase/d2q9_pf_velocity/Dynamics.c.Rt
```diff
@@ -114,8 +114,27 @@ CudaDeviceFunction vector_t calcGradPhi(){
...
 		} else if ((NodeType & NODE_BOUNDARY) == NODE_Wall){
-			gradPhi.x = (PhaseF(1,0) - PhaseF(-1,0))/3.0 + (PhaseF(1,1) - PhaseF(-1,-1) + PhaseF(1,-1) - PhaseF(-1,1))/12.0;
-			gradPhi.y = (PhaseF(0,1) - PhaseF(0,-1))/3.0 + (PhaseF(1,1) - PhaseF(-1,-1) + PhaseF(-1,1) - PhaseF(1,-1))/12.0;
+			// compute gradient with one-sided diffs but first check if neighbours are nans
+			bool nan1 = isnan(PhaseF(1,0));
+			bool nan2 = isnan(PhaseF(-1,0));
+			bool nan3 = isnan(PhaseF(0,1));
+			bool nan4 = isnan(PhaseF(0,-1));
+
+			if ((nan1 && nan2) || (!nan1 && !nan2)){
```
llaniewski, line 123: This 1. doesn't seem consistent with the previous implementation 2. gradient has different sign depending on situation 3. PhaseF are read multiple times.

I think something very consistent and robust could be achieved by some kind of a conditional average like:
```c++
gradPhi.x = 0;
gradPhi.y = 0;
phi_a = PhaseF(0,0);
for (/* all directions */) {
  phi_b = PhaseF(/* direction */);
  if (!isnan(phi_b)) {
    gradPhi.x += (phi_a - phi_b)*[/* weight */]*[/* direction x */];
    gradPhi.y += (phi_a - phi_b)*[/* weight */]*[/* direction y */];
    denom += [/* weight */];
  }
  gradPhi.x = gradPhi.x / denom;
}
gradPhi.y = gradPhi.y / denom;
```

models/multiphase/d2q9_pf_velocity/Dynamics.c.Rt
```diff
@@ -125,8 +144,27 @@ CudaDeviceFunction vector_t calcGradPhi(){
 		}
 	#else
 		if ((NodeType & NODE_BOUNDARY) == NODE_Wall){
-			gradPhi.x = (PhaseF(1,0) - PhaseF(-1,0))/3.0 + (PhaseF(1,1) - PhaseF(-1,-1) + PhaseF(1,-1) - PhaseF(-1,1))/12.0;
-			gradPhi.y = (PhaseF(0,1) - PhaseF(0,-1))/3.0 + (PhaseF(1,1) - PhaseF(-1,-1) + PhaseF(-1,1) - PhaseF(1,-1))/12.0;
+			// compute gradient with one-sided diffs but first check if neighbours are nans
```
llaniewski, line 147: Probably the `#ifdef` is in the wrong place, as this is just the same code as previously

llaniewski (changes requested):

models/multiphase/d2q9_pf_velocity/Dynamics.c.Rt
TravisMitchell: Thank you. This is a nicer method than attempting strange 1-sided diffs based on checking neighbours. The implementation is now:
`	

	gradPhi.x = 0.0;
	gradPhi.y = 0.0;

	real_t phi_a = PhaseF(0,0);
	real_t phi_b, denom = 0.0;

	for (int i = 1; i<9; i++) {
		phi_b = PhaseF_dyn(d2q9_ex[i],d2q9_ey[i]);
		if (!isnan(phi_b)){
			gradPhi.x += wf[i]*d2q9_ex[i]*(phi_b-phi_a);
			gradPhi.y += wf[i]*d2q9_ey[i]*(phi_b-phi_a);
			denom += wf[i];
		}
	}
	gradPhi.x = gradPhi.x/denom/cs2;
	gradPhi.y = gradPhi.y/denom/cs2;

`

I will do some more rigorous testing of results tomorrow and add a similar implementation for the d3q27 model :)

models/multiphase/d2q9_pf_velocity/Dynamics.c.Rt
TravisMitchell: moved ifdef for outflow conditions after previous calc to remove repeated code.

PR #479, opened 2024-01-04
Corrections for uninitialised fields in d3q27_pf_velocity

llaniewski:
This fixes the problem of uninitialised values in the `InitFromFieldsStage` stage.

@shkodm Could you make it a priority to test if this doesn't break your computations?


shkodm:
It seems that breaks something... I am investigating now

shkodm:
> It seems that breaks something... I am investigating now

Never mind, everything seems to be working fine

shkodm (approved):

PR #487, opened 2024-01-10
Found on bunya make fails (unless nlopt disabled) due to mpi.rank

TravisMitchell:
I think this was just an outdated reference with mpi.rank - switched to mpi_rank as per other handlers and compile goes ahead.


llaniewski:
@TravisMitchell This has a bit messy commit history, so I'll "squash" it. After that you can replace your develop branch with this.

PR #493, opened 2024-01-15
Add fixes for building on macOS

shkodm:
- `-E` in sed to be POSIX compliant
- Remove regex in `find` to make it more tolerable and portable
- `find models/` produces extra slashes in results on macos ( i.e. `models//multiphase/`) but not on Linux. `find models` does not give extra slashes and behaves the same as `find models/` on models

Tested on my M2 Macbook Air, all the clusters where I use TCLB, and my workstation. 

@TravisMitchell  FYI 

I will also CI at some point, but not in this PR.


shkodm:
Superseded by https://github.com/CFD-GO/TCLB/pull/495

PR #495, opened 2024-01-15
Add fixes for building on macOS

shkodm:
Supersedes #493

- `-E` in sed to be POSIX compliant
- Remove regex in `find` to make it more tolerable and portable
- `find models/` produces extra slashes in results on macos ( i.e. `models//multiphase/`) but not on Linux. `find models` does not give extra slashes and behaves the same as `find models/` on models

Tested on my M2 Macbook Air, all the clusters where I use TCLB, and my workstation. 

@TravisMitchell  FYI 

I will also CI at some point, but not in this PR.

### DIFF
--- a/src/makefile.main.Rt
+++ b/src/makefile.main.Rt
@@ -293,7 +293,7 @@ wiki/schema/<?%s m ?>.xsd:$(SRC)/schema.xsd.Rt $(SRC)/conf.R <?%s d ?>/<?%s m ?>
 %_sp.c:%.c
 	@echo "  SP-CONST   $@"
 	@$(MKPATH) $@
-	@cat $< | sed -r 's/([^a-zA-Z0-9][0-9]+\.[0-9]*([eE][-+]?[0-9]+)?)[fL]+/\1f/g' > $@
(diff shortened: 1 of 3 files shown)


llaniewski:
Could you verify also if things that are recommended for install here:
https://github.com/CFD-GO/TCLB/blob/master/tools/install.sh#L292
are in tune with what you have?

Also, I remember other people running TCLB on mac needed some paths changed, and back then I wrote this:
https://github.com/CFD-GO/TCLB/blob/master/tools/env.sh
Is this still needed?

shkodm:
> Also, I remember other people running TCLB on mac needed some paths changed, and back then I wrote this:
https://github.com/CFD-GO/TCLB/blob/master/tools/env.sh
Is this still needed?

I did not use it. Both in the script and in the `tools/install.sh`,  `gnu-sed` and `findutils` are not needed, because macOS already ships those tools by default, and it is better not have random conflicts between them

> Could you verify also if things that are recommended for install here:
https://github.com/CFD-GO/TCLB/blob/master/tools/install.sh#L292
are in tune with what you have?

I installed using the `install.sh` (and have all the packages that are installed there), but taking into the account the notice above. I don't think I did anything extra. When @TravisMitchell is going to install, he can verify.

TravisMitchell:
Thanks Dima, it will be a couple weeks before I get the chance to test (lead time is on my mac is a bit slow) - but will let you know if anything else is needed for this.

PR #515, opened 2024-05-27
Fixed to the phase-field model

shkodm:
Synchronising various fixes I did to the phase-field mode over the past months:

- Split initialisation of wall normals in 2 parts, so it is possible to supply user defined wall normals using RunR 
- Fix various edge cases and bugs with wetting boundary conditions
- Fix pressure boundary conditions

Pressure boundary conditions fixes: 
- Now use proper ZouHe boundary conditions, with 3 variants.  (Some can be removed later)
  - Standard (e.g. `EPressure`) makes sure the inlet / outlet phase-field is the same as specified (i.e. sum of distributions forced to be equal to it). 
  - `open` suffix (i.e. EPressureopen) does not enforce the phase-field
  - `fflux` enforces constant flux
 - Gradient / Laplacian on the velocity / pressure boundaries is always zero (does not touch the other side of the domain)
 - Wall normal calculations do not take into account the other side of the domain (periodicity is switched off). This works only for the case of X direction and when the flag `InvasionDrainage` is non-zero. And when the boundary conditions are the first / last layers of the domain. Otherwise, one has to compile the model with `autosym` option, and mark the inlet / oulets with Symmetry boundaries, so that the periodicity is switched off 
 - `InvasionDrainage` marks if the case is drainage (non-wetting displacing wetting), or invasion (wetting displacing non-wetting). This sets some densities and phase-field on the inlet / outlet
 
 Merging with master, because I (and I think other people who are using phase-field model are mostly using master branch and the fixes are quite crucial)

Now doing some testing to make sure I did not mess up when refactoring. So draft PR for now. 
 cc @TravisMitchell

### DIFF
--- a/models/multiphase/d3q27_pf_velocity/Boundary.c.Rt
+++ b/models/multiphase/d3q27_pf_velocity/Boundary.c.Rt
@@ -735,8 +735,9 @@ CudaDeviceFunction void calcWallPhase(){
 			<?R } ?>
 
 
-			// dont do staircase improvement if any of the interpolating nodes are solid
-			if (IsSpecialBoundaryPoint != <?%d NEXT_INTERPOLATING_POINTS_ARE_SOLID ?>) {
+			// dont do staircase improvement if any of the interpolating nodes are solid, make sure to cover
+			// both cases......
+			if (IsSpecialBoundaryPoint != <?%d NEXT_INTERPOLATING_POINTS_ARE_SOLID ?> && IsSpecialBoundaryPoint != <?%d BOTH_NEXT_AND_NNEXT_INTERPOLATING_POINTS_ARE_SOLID ?>) {
 				real_t pf_interpolated = coeff_v1 * pf_v1 + coeff_v2 * pf_v2 + coeff_v3 * pf_v3;
 				h = 0.5 * sqrt(nw_actual_x*nw_actual_x + nw_actual_y*nw_actual_y + nw_actual_z*nw_actual_z);
 				pf_f = pf_interpolated;
--- a/models/multiphase/d3q27_pf_velocity/Boundary.c.Rt
+++ b/models/multiphase/d3q27_pf_velocity/Boundary.c.Rt
@@ -70,55 +70,72 @@ CudaDeviceFunction void <?R C(my_velocity_boundaries[ii])?>(){
 		C(h[sel], heq[sel] + heq[bounce][sel] - h[bounce][sel] - 0.5 * Unknowns * (exM))
 	?>
 }
+<?R
+}
+?>
 
-CudaDeviceFunction void <?R C(my_pressure_boundaries[ii])?>(){
+<?R
+for (ii in 1:cube_faces){
+for (j in 1:length(pressure_boundary_types)) {  ?>
+CudaDeviceFunction void <?%s paste0(C(my_pressure_boundaries[ii]), pressure_boundary_types[j]) ?>(){
     real_t d = getRho();
-
     real_t myPhaseField = 0.0;
-    // TODO: Need a proper fix soon
-    if ((NodeType & NODE_BOUNDARY) == NODE_WPressure){
-        d = Density_l;
-        myPhaseField = 0.0;
+
+    if ((NodeType & NODE_BOUNDARY) == <?%s paste0("NODE_WPressure", pressure_boundary_types[j]) ?>){
+        d = InvasionDrainage == 2? Density_l : Density_h;
+        myPhaseField = InvasionDrainage == 2? 0.0 : 1.0;
     }
 
-    if ((NodeType & NODE_BOUNDARY) == NODE_EPressure){
-        d = Density_h;
-        myPhaseField = 1.0;
+    if ((NodeType & NODE_BOUNDARY) == <?%s paste0("NODE_EPressure", pressure_boundary_types[j]) ?>){
+       d = InvasionDrainage == 2? Density_h : Density_l;
+       myPhaseField = InvasionDrainage == 2? 1.0 : 0.0;
     }
 
 	real_t pstar = Pressure / (d*cs2);
-
     <?R
-    type_h = "fix_flux"
+    type_h = (if (j == 1) "fpf" else pressure_boundary_types[j])
     type = "pressure"
     n = my_normal_directions[ii,]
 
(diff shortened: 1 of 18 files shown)


models/multiphase/d3q27_pf_velocity/Boundary.c.Rt
```diff
@@ -70,42 +70,127 @@ CudaDeviceFunction void <?R C(my_velocity_boundaries[ii])?>(){
...
+<?R
+}
+?>
 
-CudaDeviceFunction void <?R C(my_pressure_boundaries[ii])?>(){
+<?R
+for (ii in 1:cube_faces){
+for (j in 1:length(pressure_boundary_types)) {  ?>
+CudaDeviceFunction void <?%s paste0(C(my_pressure_boundaries[ii]), pressure_boundary_types[j]) ?>(){
     real_t d = getRho();
```
TravisMitchell, line 81: @shkodm should d be based off a prescribed phase on the boundary rather than from getRho - which I believe calls from PhaseF field

shkodm: yea, it is then set properly in the if condition below

TravisMitchell: I did see this, but figured it was maybe not best to define it with a value knowing it will be overwritten. I was also curious how this worked with the open condition - but perhaps I need to just look at the code a bit deeper. I will try to do this before tomorrow when we meet but have a bit of a marking pile with end of semester.

models/multiphase/d3q27_pf_velocity/Boundary.c.Rt
```diff
@@ -845,12 +963,89 @@ CudaDeviceFunction void calcWallPhase(){
...
+  }
 }
 
 CudaDeviceFunction real_t getSpecialBoundaryPoint() {
 	return IsSpecialBoundaryPoint;
 }
 
+/* Calculate the actual wall norm based on the solid
+ *field */
+CudaDeviceFunction real_t Init_real_wallNorm()
```
llaniewski, line 975: This function should be void (shouldn't be returning anything).

shkodm:
@nparaskevas you can test it now if you want. I provided the description of different boundary conditions variants in the PR description. Just using `WPressure` on the inlet and `EPressurefflux` or `EPressurefpf` or `EPressureopen` on the outlet would work, but you should play around. 

Don't forget to add `InvasionDrainage value="2"` to the setting in case of drainage, or `value = 1` in case of invasion

models/multiphase/d3q27_pf_velocity/Boundary.c.Rt
```diff
@@ -70,42 +70,131 @@ CudaDeviceFunction void <?R C(my_velocity_boundaries[ii])?>(){
...
-		U_PF = U[1:PF_velocities,]
-       	pf = PV("PhaseField")
-		n = my_normal_directions[ii,]
-		heq = pf * EQ_h$feq
-		bounce = Bounce(U_PF)
-		sel = as.vector( (U_PF %*% n) < 0)
+    if ((NodeType & NODE_BOUNDARY) == <?%s paste0("NODE_EPressure", pressure_boundary_types[j]) ?>){
+       d = InvasionDrainage == 2? Density_h : Density_l;
+       myPhaseField = InvasionDrainage == 2? 1.0 : 0.0;
+    }
```
TravisMitchell, line 93: @shkodm I think here you should be setting:
`
PhaseF = PhaseField
d = getRho()
`
I'm not sure what your InvasionDrainage setting here is checking as you specify it should be 1 or -1 in the Dynamics.R. This is causing an issue (i think) for pressure pressure cases where fpf is used as it effectively forces PhaseField=0 on all pressure boundaries.

shkodm: @TravisMitchell The documentation in `Dynamics.R` is incorrect (copypasted), `InvasionDrainage` should have values 1 or 2. `d` should be set explicitly because I don't want any streamed values obtained from `getRho()` This was the whole point of those if statements, without it would not work properly.

TravisMitchell: @shkodm, wouldn't it be much simpler/neater to take `PhaseField` (zonal setting defined in xml) and then interpolate for `d` based off `Density_h` and `Density_l`. This would mean that the `myPhaseField` would be simply set from the zonal condition applied in the xml rather than relying on this `InvasionDrainage` setting.

shkodm: @TravisMitchell yea, interpolation is a good idea. I think I implemented it with an additional setting originally, because `InvasionDrainage` (or whatever the name of the flag) is still necessary to  know if we want to "reflect/mirror" the boundaries on the inlet (to have proper wall normal not pointing to the other side of domain), or if we want to keep periodic boundaries. 

Anyway, I will remove the if statements here later today or tomorrow and double check the boundary conditions.

llaniewski:
@shkodm @TravisMitchell What's the status on this?

TravisMitchell:
> @shkodm @TravisMitchell What's the status on this?

@shkodm - can you have a quick look through the old requested changes and confirm if this is ready for merging?

For special cells we should look at [this method](https://doi.org/10.1016/j.jcp.2013.05.012) to check if it would be a better option than averaging but I think this can wait till a later stage. The priority should be getting the fixes here merged.

shkodm:
New updates (based on the stable version from my working branch): 
- Fixes old required changes (mainly just reorganising table for the Boundary conditions, as per Lukaszes comment). It is indeed cleaner now
- Rewrote the "special boundary point" handling to weighted averaging all the time (more stable)
- Added saturation tracking 
- Added velocity limiter (`vlimiter > 0` enables it based on `vlimiter` value)
- Added simple tanh initialisation of the interface at point

@llaniewski @TravisMitchell can you check if you are happy with the structure now? I don't want to add further things, there are already quite a lot of changes

TravisMitchell:
@llaniewski is there any other mods you want on this pull request -- it would be great if we could get this merged so that future changes can build from this. 

The TCLB test fails due to change in globals which means the csvdiff doesn't pass, which will require an update to TCLB_tests.

PR #516, opened 2024-05-30
Fixing the globals in RunR

llaniewski:
Adding `ITER_LASTGLOB` to RunR element, and to the last iteration of Solve action.

@shkodm, could you test this?


shkodm (approved):
Tested for several use cases (RunR element that is called periodically, and RunR element executing once after Solve element, RunR element together with Log). All works smoothly :+1:

PR #527, opened 2025-09-02
Handler Average extend to arbitrary lattice

lisataldir:
I added a handler <Average/> for arbitrary lattice

### DIFF
--- a/src/ArbLattice.cpp.Rt
+++ b/src/ArbLattice.cpp.Rt
@@ -1,3 +1,8 @@
+<?R
+    source("conf.R")
+    c_header();
+?>
+
 #include "ArbLattice.hpp"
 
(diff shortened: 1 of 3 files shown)


llaniewski:
Hi, @lisataldir,

I would like to ask you to redo this PR, not promoting code to R-templates. The necessary information you need (to loop over fields and select the ones to average) is provided in the model structure (see https://github.com/CFD-GO/TCLB/blob/develop/src/Lists.cpp.Rt), so you can make this loop in C++. You can see an example here: https://github.com/CFD-GO/TCLB/blob/c6b1675d10b1b52fcb344849690248ad086d945f/src/CartLattice.cpp.Rt#L629

Best,

PR #528, opened 2025-09-02
Feature/handlers

lisataldir:
I modified the resetAverage function to use a C++ loop instead of R-template


llaniewski:
Sorry to be a spoilsport, but you left the previous commit and the `.Rt` files.

Make a new branch from develop, introduce the change and make a clean pull request.

PR #537, opened 2025-12-09
Feature electricfield

engparth-cmd:
I added the mode ...
I modified the model ... (mention the original author with @...)
I added a handler ...
I fixed ... (mention issues with #...)
I added needed references in ...


TravisMitchell:
@engparth-cmd, was this an accidental pull request? It does not look like it should be merged into Master, which is a stable branch.

PR #538, opened 2025-12-19
Sample for ArbLattice new version

lisataldir:
I made some changes on the handler <Sample/> for Arbitrary lattice following the last recommendations


llaniewski:
Done in #541

PR #540, opened 2026-03-05
Fixes to particle coupling on arbitrary grid

shkodm:
- Added subgrid particles for the cumulant model (already tested by some working students in previous years, existed somewhere on my branch), useful to have it in the main repo and not free floating somewhere among my branches
- Elevated getLocalBoundingBox to the LatticeBase method. Also noticed that ArbitraryLattice does not have options for `px`, `py`, `pz` offsets
- Added destructor the the arbitrary lattice (other RFI connection never closes, which leads to some nice 100K lines csv files...) 
- **Important bug fix**: Previously X, Y, Z coordinates (as accessible from the node, or when using RInside for instance), were inconsistent between cartesian grid and arbitrary grid. Cartesian grid would always return them in lattice units (which is correct, especially for the particle coupling), whilst arbitrary grid returns them in real units (because they are written as real units to the `cxn` file). So if units were sets, arbitrary grid would have trouble finding particles correctly. And the results from RInside were completely wrong - multiplied by units twice. I changed that to store coordinates in LBM units, and convert using the grid spacing when necessary. 

Tested by: 
- Running locally for general functionality, testing unit conversion and checking results from RInside
- Running a simple case with simplepart on Setonix with and without arb grid and comparing particle coordinates -> match to 1-e10 precision.

### DIFF
--- a/models/flow/d3q27_cumulant_part/Dynamics.c.Rt
+++ b/models/flow/d3q27_cumulant_part/Dynamics.c.Rt
@@ -20,34 +20,78 @@
 #	EQ = MRT_eq(U, rho, J);
 ?>
 
+#define PI 3.141592653589793
+
 CudaDeviceFunction void CalcF(){
 	for (auto p : SyncParticleIterator(X,Y,Z)) {
-	if ((NodeType & NODE_BOUNDARY) == 0) {
-	real_t d = getRho();
+        real_t d = getRho();
         vector_t u;
-	vector_t force;
-<?R C(PV(c("u.x","u.y", "u.z")), f %*% U) ?>
-	if (p.dist < p.rad) {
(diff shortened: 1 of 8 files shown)


llaniewski (changes requested):
@shkodm, this is good. Just fix the `override` tags the clang compiler is failing on https://github.com/CFD-GO/TCLB/actions/runs/22888346758/job/66406017687?pr=540#step:5:246

shkodm:
~I think in some bizzare way missing overrides were not flagged by compilers until I removed the default destructor..~  The error because the class interface was not consistent, with only one function marked as override. I added overrides to all functions, so it consistent now (in CartLattice for instance we have all overrides marked)

shkodm:
@llaniewski this should be good now

llaniewski (approved):

PR #541, opened 2026-03-05
Implemented sampler on Arbitrary grid and fixes for cartesian one

shkodm:
## On Arbitrary grid: 
- Moved sample to the lattice base
 - Implemented executor to map cartesian coordinate into local id by an array scan with a tolerance. This is the only solution I think that avoid additional array allocations
 - Otherwise the functions are similar to the cartesian ones
 - To not search for index on every iteration, cached the point index inside the sample points attribute (not that much extra memory, should be ok) 
 ## On Cartesian grid
 - Fixed a bug that on multiple GPUs the sample point location was never converted to the local coordinate frame, so was returning wrong results (because `intersect` does the intersection of bounding boxes, not the conversion into coordinate frames). Fix by offsetting the point into the local coordinate system 
 
 Tested by: 
 - Print local ids of different sampling points across 1 and 2 gpus 
 - Running sampler with several points on 1 GPU, and 2 GPUs and comparing results for both cartesian and arb lattice (to make sure the distribution of points works) - this is how I found the problem with the cartesian lattice, I think Lisa saw something like that previously 
 - Comparing sampler results between cartesian and arbitrary lattice on 1 and 2 gpus - no difference
 
 Bounding Box changes are the same as in #540

### DIFF
--- a/src/ArbLatticeLauncher.hpp.Rt
+++ b/src/ArbLatticeLauncher.hpp.Rt
@@ -153,7 +153,6 @@ struct CartesianCoordinateMapperArbExecutor: public LinearExecutor {
 
     CudaDeviceFunction void Execute() const {
         using LA = ArbLatticeAccess; 
-        using N = Node< LA, Primal, NoGlobals, Get >; 
         const int i = threadID(CudaThread, CudaBlock, CudaNumberOfThreads);
         if (inRange(i)) {
             LA acc(i, container);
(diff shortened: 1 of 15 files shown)


src/ArbLatticeLauncher.hpp.Rt
```diff
@@ -129,10 +132,69 @@ if (q$type == "vector_t") {
...
+struct CartesianCoordinateMapperArbExecutor: public LinearExecutor {
+    ArbLatticeContainer container; 
+    LatticeData data;
+    vector_t point; // cartesian point coordinates
+    real_t epsilon2;  // should be a small value (e.g. < 1e-10) to match with only one point
+    unsigned int *matched_lid; 
+
+    CudaDeviceFunction void Execute() const {
+        using LA = ArbLatticeAccess; 
+        using N = Node< LA, Primal, NoGlobals, Get >;
```
llaniewski: No node type is used here.

llaniewski (changes requested):

shkodm:
Had the same issue with override virtual functions as in https://github.com/CFD-GO/TCLB/pull/540#issuecomment-4044226691. Should be fixed now

llaniewski (approved):

PR #542, opened 2026-03-05
Implement cuts for the arbitrary lattice

shkodm:
Mostly transposing changes from the Lisa's branch. Adjusted for conditional reading / writing / allocating of cut distances when nothing is cut


llaniewski (approved):
Great work @lisataldir @shkodm

PR #543, opened 2026-03-09
Fixes to  RunR for the arbitrary grid

shkodm:
Mostly added getters and setters.

Also fixed a compilation error for the phase field, when compiling with >= c++17 .

Should be noted that for instance for scalar quantities, arbitrary grid returns 1D arrays with respect to the geometry,
whereas cartesian returns 3D arrays. For complex setup one can just use `Solver$GEOMETRY$BOUNDARY` and
`Solver$Geometry$X`(or `Y`, `Z`) arrays, instead of assuming certain mesh structure. 

Tested by:
- Running interactive RunR session for cartesian and arbitrary grid
- Initialising droplet inside a periodic domain using RunR

### DIFF
--- a/src/ArbLattice.cpp
+++ b/src/ArbLattice.cpp
@@ -517,13 +517,41 @@ void ArbLattice::SetFirstTabs(int tab_in, int tab_out) {
 }
 
 
+std::vector<big_flag_t> ArbLattice::getFlags() const {
+    size_t size = getLocalSize();
+    std::vector<big_flag_t> ret(size);
+    CudaMemcpy(ret.data(), launcher.container.node_types, sizeof(big_flag_t)*getLocalSize(), CudaMemcpyDeviceToHost);
+    return ret;
+}
 
+std::vector<real_t> ArbLattice::getField(const Model::Field& f) {
+	assert(f.isParameter);
+    storage_t* from = getSnapPtr(Snap);
+	assert(sizeof(storage_t) == sizeof(real_t));
+    size_t size = getLocalSize();
+    std::vector<real_t> ret(size);
+	debug2("Pulling all %s\n",f.name.c_str());
+    CudaMemcpy(ret.data(), &from[f.id*sizes.snaps_pitch], size*sizeof(real_t),
+               CudaMemcpyDeviceToHost);
+    return ret;
+}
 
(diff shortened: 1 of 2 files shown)


src/ArbLattice.cpp
```diff
@@ -517,13 +517,41 @@ void ArbLattice::SetFirstTabs(int tab_in, int tab_out) {
...
+               CudaMemcpyDeviceToHost);
+    return ret;
+}
 
-std::vector<big_flag_t> ArbLattice::getFlags() const { throw std::runtime_error{"UNIMPLEMENTED"}; return {}; };
-std::vector<real_t> ArbLattice::getField(const Model::Field& f) { throw std::runtime_error{"UNIMPLEMENTED"}; return {}; };
 std::vector<real_t> ArbLattice::getFieldAdj(const Model::Field& f) { throw std::runtime_error{"UNIMPLEMENTED"}; return {}; };
-void ArbLattice::setFlags(const std::vector<big_flag_t>& x) { throw std::runtime_error{"UNIMPLEMENTED"}; return; };
-void ArbLattice::setField(const Model::Field& f, const std::vector<real_t>& x) { throw std::runtime_error{"UNIMPLEMENTED"}; return; };
+void ArbLattice::setFlags(const std::vector<big_flag_t>& x) { throw std::runtime_error{"NOT SUPPORTED FOR ARB LATTICE"}; return; };
```
llaniewski: Please add `setFlags` analogous to `getFlags`.

llaniewski (changes requested):
This is good to merge, but I would like the `setFlags` for completeness.

src/ArbLattice.cpp
shkodm: @llaniewski  I did not implement it, because this would allow overwriting the `Wall` markers on nodes. Do we really want it for the arbitrary grid? I think it would the user should not try to do it

llaniewski:
Yeah, the user can overwrite any flags he wants. Not always will it make
sense, but no reason to block it.

shkodm:
@llaniewski the failing check is some github action problems (could not check out the repo). I think this is good to merge

llaniewski:
I've rerun the action and all checks are good.

llaniewski (approved):

PR #544, opened 2026-03-09
Supressing error on wrongfully checked allocation sizes in new gcc

llaniewski:


llaniewski:
@shkodm Can you update or rebase your branches with this so we know if all the CI errors are related to this?

